PR #11, opened 2020-11-25
Remove unused variable in RecordBuilder

TedCassirer:
Minor issue
There's an unused variable `r` in the builder if the record only have 1 field.
Spotbugs flagged the generated source in my project because of it. This should fix the issue.

Example:
```
@RecordBuilder
public record MyRecord(long id) implements MyRecordBuilder.With {}
```
Generates ->
```
public class MyRecordBuilder {
        private long id;

...
....
        /**
         * Return a new instance of {@code MyRecord} with a new value for {@code id}
         */
        default MyRecord withId(long id) {
            var r = (MyRecord)(Object)this;  //<---UNUSED
            return new MyRecord(id);
        }
}
```

I could just exclude the generated sources from spotbugs if you consider this a non bug

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -188,9 +188,11 @@ class InternalRecordBuilderProcessor
                 return new MyRecord(name, r.age());
             }
          */
-        var codeBlockBuilder = CodeBlock.builder()
-                .add("$T r = $L(this);\n", recordClassType.typeName(), metaData.downCastMethodName())
-                .add("return new $T(", recordClassType.typeName());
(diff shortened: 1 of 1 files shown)


Randgalt:
Thanks again.

TedCassirer:
Ah, I commited a misstake. Noticed it immediately after creating the PR. Sorry!
I'll push the correct fix in a min

Randgalt:
lol - I should test these before I merge

TedCassirer:
Disregard that. I just messed up updating my dependencies in my project causing me to build with the incorrect fix. It works as intended.
```
        /**
         * Return a new instance of {@code MyRecord} with a new value for {@code id}
         */
        default MyRecord withId(long id) {
            return new MyRecord(id);
        }
```

Randgalt:
I apologize, Github isn't giving you credit for these PRs due to: https://github.com/isaacs/github/issues/1368

Issue #15, opened 2020-12-11
ElementUtils relies on toString() to get package name


mbarbero:
Depending on the implementation of `javax.lang.model.element.Element`, the `toString` method may return different values.

It has been discovered while trying to execute the annotation processor within the Java support for vscode (https://github.com/eclipse/eclipse.jdt.ls/issues/1629) where the `Element` class is implemented by the Eclipse JDT compiler.

I'll provide a PR with a fix shortly.

Randgalt:
Fixed by #16

PR #16, opened 2020-12-11
Stop relying on toString() to detect package name.

mbarbero:
Fix for #15.

Credit goes to @snjeza for finding this out as part of https://github.com/eclipse/eclipse.jdt.ls/issues/1629

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/ElementUtils.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/ElementUtils.java
@@ -86,8 +86,9 @@ public class ElementUtils {
                 break;
             }
         }
-        String name = typeElement.getEnclosingElement().toString();
-        return !name.equals("unnamed package") ? name : "";
+        String name = typeElement.getQualifiedName().toString();
+        int index = name.lastIndexOf(".");
+        return (index > -1) ? name.substring(0, index) : "";
     }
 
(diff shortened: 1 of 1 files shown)


record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/ElementUtils.java
```diff
@@ -86,8 +86,9 @@ public static String getPackageName(TypeElement typeElement) {
                 break;
             }
         }
-        String name = typeElement.getEnclosingElement().toString();
-        return !name.equals("unnamed package") ? name : "";
+        String name = typeElement.getQualifiedName().toString();
```
Randgalt, line 89: We lost the check for "unnamed package" - is that no longer needed?

mbarbero: Indeed, it's no longer required: the qualified name of a *typeElement* is the only name of the element when inside the default/unnamed package while it's the fully qualified name (dot separated) of the package + name of the *typeElement* otherwise (e.g. `a.b.c.MyClass`). 

The dot separated qualified name is part of the contract of the `javax.lang.model.element.Element` class, contrary to the output of the `toString()` method.

Randgalt:
Dammit - it used my email when merging. I apologize. This has been a problem on Github

mbarbero:
> Thanks for the contribution. I'll build a release soon.

Anytime. Thanks for the quick turnaround.

PR #17, opened 2020-12-21
Use compileOnly instead of implementation

marcphilipp:
Since most users only need the annotations.


Randgalt:
Let's include Maven too: https://github.com/Randgalt/record-builder/pull/18

marcphilipp:
Superseded by #18

PR #38, opened 2021-06-20
Set optional fields to empty by default

### LINKED ISSUE #34: Default values in builder
Currently the fields of a generated builder are initialized with Java default values. Thus if one had a field of type `Optional<SomeFancyClass>` its initial value would be `null` whereas `Optional.empty()` may be a preferable initial value.

Randgalt:
Sets `Optional` fields to `empty()` by default. Adds an option
to control this.

Closes #34

### DIFF
--- a/java16.sh
+++ b/java16.sh
@@ -16,6 +16,6 @@
 #
 
 
-jenv local 16-ea
+jenv local 16
 javahome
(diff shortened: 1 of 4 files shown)


Randgalt:
cc @TrackerSB

PR #48, opened 2021-06-30
Validation and null checks are missing for withers. This PR adds them.

### LINKED ISSUE #47: Validations for the withers
Hi Jordan

I'm playing around with new validation opportunities that you just have implemented. Very nice :-) 

Howevere, It seems to be a possible to sneak thru the validation when using the withers methods.

For example when letting 
RequiredRecord implement RequiredRecordBuilder.With
and 
RequiredRecord2 implement RequiredRecord2Builder.With 

I expect these two unit test to pass:

``` @Test
    void testNotNullsWithNewProperty() {
        var valid = RequiredRecordBuilder.builder().hey("hey").i(1).build();
        Assertions.assertThrows(NullPointerException.class, () -> valid.withHey(null));
    }

    @Test
    void testValidationWithNewProperty() {
        var valid = RequiredRecord2Builder.builder().hey("hey").i(1).build();
        Assertions.assertThrows(ValidationException.class, () -> valid.withHey(null));
    }
```
For the @RecordBuilder.Options(interpretNotNulls = true) 
I guess it just as easy to add Objects.reuireNonNull in appropriate with-methods 

but how to handle @RecordBuilder.Options(useValidationApi = true) ?

We dont want to perform a full validation of the entiere object via the builders static "new" method
but instead only validate the actual property. 

Please let me know your thoughts around this and I'll be happy to start with an initial PR.

Regards
Dan

Randgalt:
Closes #47

cc @danp11

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -206,10 +206,15 @@ class InternalRecordBuilderProcessor {
             }
          */
         var codeBlockBuilder = CodeBlock.builder();
+        addNullCheckCodeBlock(codeBlockBuilder, index);
         if (recordComponents.size() > 1) {
             codeBlockBuilder.add("$T $L = $L(this);\n", recordClassType.typeName(), uniqueVarName, metaData.downCastMethodName());
         }
-        codeBlockBuilder.add("return new $T(", recordClassType.typeName());
+        codeBlockBuilder.add("$[return ");
+        if (metaData.useValidationApi()) {
+            codeBlockBuilder.add("$T.validate(", validatorTypeName);
+        }
+        codeBlockBuilder.add("new $T(", recordClassType.typeName());
         IntStream.range(0, recordComponents.size()).forEach(parameterIndex -> {
             if (parameterIndex > 0) {
                 codeBlockBuilder.add(", ");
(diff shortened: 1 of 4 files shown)


danp11:
Hi @Randgalt and thanks for making this PR.

When I'm looking at the generated code I have to questions/concerns:

```default RequiredRecord2 withHey(@NotNull String hey) {
            RequiredRecord2 r = RequiredRecord2Builder._downcast(this);
            return (RequiredRecord2)RecordBuilderValidator.validate(new RequiredRecord2(hey, r.i()));
        }

        default RequiredRecord2 withI(@NotNull int i) {
            RequiredRecord2 r = RequiredRecord2Builder._downcast(this);
            return (RequiredRecord2)RecordBuilderValidator.validate(new RequiredRecord2(r.hey(), i));
        }```

1) Eventhough records are recommended to use "as DTO's and Value types" I believe that we in the future will work with large records with any variables. The current generated code will therefore, IMO, have a svere impact on performance if the validation has to be done on all the properties even if we just want to use a with-method on one (1) property.

2) If the parameter is a primitive, does it really make sence to do an extra validation then?

Just my initial thoughts, 

Regards
Dan

danp11:
Wouldn´t it be appropriated to do the Jsr-303 validation on a property level instead as you suggested here:

```Yes, I agree. For validation, you can validate individual fields (referenced as "properties" in the spec) via: https://stackoverflow.com/a/21961677/2048051. A PR would be appreciated.```

Thanks.

Randgalt:
@danp11 I started using property validation and quickly realized that that doesn't work. What if there are custom validations that assert `@Min` and `@Max` based on other fields of the record? So, you must validate the entire record every time as any single change might affect other validations. The `With` has methods to change multiple fields at once. It's assumed if a single `withXXX()` is called the intent is to change only 1 component otherwise one of the multi-field with methods would be used.

danp11:
@Randgalt Ah, yes, you're correct. 

Then the code additions in this PR lgtm.

PR #49, opened 2021-07-16
Remove downcast in favor of methods

### LINKED ISSUE #27: Sealed wither interface?
Its still down the line (Java 17, most likely), but it would be nice to have the `RecordName.With` generated interface be sealed so that only the record can implement it

Randgalt:
Great suggestion from @Twisol. There's no need for the downcasting
if record component methods are added to the Wither interface.

Closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,29 +184,26 @@ public class NameAndAgeBuilder {
                 && Objects.equals(name, b.name)
                 && (age == b.age));
     }
-
-    /**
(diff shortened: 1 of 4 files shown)


grahamrvvup:
@Randgalt lgtm, that's a neater solution and the version works in our build without warnings 👍

Randgalt:
OK - I'll merge this and create a release soon.

grahamrvvup:
ty 🙏

PR #60, opened 2021-08-13
Adds special handling for collection record components

### LINKED ISSUE #56: Options to create unmodifiableLXYZ
Hi again:-) 

Just thinking if it could be a good idea to have an new RecordBuilder.Options
like "useUnmodifiableCollections" or similar? 

And then for List/Map/Set use Collections.unmodifiableXYZ in the builder..?

What is your thoughts around this suggestion?

Have a nice weekend :-) 

/Dan

### LINKED ISSUE #58: Add emptyDefaultForCollections options
I think many people would like to have `emptyDefaultForCollections`, where instead of nulls builder would set following instances in the record built:

* List -> List.of()
* Set -> Set.of()
* Map -> Map.of()
* Collection -> Set.of() or List.of()

Randgalt:
Adds special handling for record components of type `java.util.List`, `java.util.Set`, `java.util.Map` and `java.util.Collection`. When the record is built, any components of these types are passed through an added shim method that uses the corresponding immutable collection (e.g. `List.copyOf(o)`) or an empty immutable collection if the component is null.

This is what the generated builder looks like: [CollectionRecordBuilder.java](https://gist.github.com/Randgalt/01acd70fedb2c474b5af6d8d0fc452a8)

Closes #56
Closes #58

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -136,6 +136,14 @@ public @interface RecordBuilder {
          * named {@code RecordBuilderValidator} which has a public static method: {@code public static <T> T validate(T o)}.</p>
          */
         boolean useValidationApi() default false;
+
+        /**
+         * Adds special handling for record components of type {@link java.util.List}, {@link java.util.Set},
+         * {@link java.util.Map} and {@link java.util.Collection}. When the record is built, any components
+         * of these types are passed through an added shim method that uses the corresponding immutable collection
+         * (e.g. {@code List.copyOf(o)}) or an empty immutable collection if the component is {@code null}.
+         */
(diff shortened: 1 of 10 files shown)


Randgalt:
@krzyk I believe this satisfies your request. I didn't see the point in only adding an initializer for collections.

Randgalt:
cc @danp11

danp11:
Very nice and Thanks for adding this!

Only thing Im thinking of is if it should be mentioned in RecordBuilder and in the JavaDoc for useImmutableCollections that @NotNull and interpretNotNullsPattern = true takes precedence over "or an empty immutable collection if the component is {@code null}."

Or maybe it is better to leave that out to not complicate things to much :-) 

Thanks again for this PR.

Randgalt:
@danp11 yeah - I think letting it alone is best. Maybe it can be improved in the future, who knows. If you use a combination of options maybe a different generation mode emerges.

PR #71, opened 2021-09-22
Optionally include all records from a list of packages

### LINKED ISSUE #67: Generate builders for all classes in package
### Discussed in https://github.com/Randgalt/record-builder/discussions/66

<div type='discussions-op-text'>

<sup>Originally posted by **nikolavojicic** August 30, 2021</sup>
``` java
@RecordBuilder.Include({
    ImportedRecord1.class,
    ImportedRecord2.class,
})
public class Config {}
```
Is it possible to pass package path(s) somehow, instead of classes, so that builders for all classes in those packages are generated?</div>

Randgalt:
Closes #67

cc: @nikolavojicic

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,9 @@ public void Placeholder {
 }
 ```
 
+`@RecordBuilder.Include` also supports a `packages` attribute that includes all records
+in the listed packages.
+
 The target package for generation is the same as the package that contains the "Include"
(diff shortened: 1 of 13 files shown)


Randgalt:
@nikolavojicic can you test with this? Is this a reasonable implementation?

nikolavojicic:
Sorry for the late response.  
Yes, that is what I needed, thx!

PR #74, opened 2021-09-27
Support single item collection builders

### LINKED ISSUE #73: Generate methods in builder to add a single value to collections
First of all thanks for making this library. Awesome project. 

I'm in the middle of upgrading a project that heavily uses Lombok. We use [`@Value`](https://projectlombok.org/features/Value) + [`@Builder`](https://projectlombok.org/features/Builder) a lot so the combination of Java records and `@RecordBuilder` is a really nice upgrade path. 

Lombok provides a way to generate "singular" builder methods using `@Singular` annotation on a field. It would be really nice if the builder generated with `@RecordBuilder` would include these by default. 

[Immutables](https://immutables.github.io/immutable.html#array-collection-and-map-attributes) also generate builder methods to add single items to collections.

An example:

```java
public record Foo(List<Item> items) {}
```

The generated builder would include

```java
// this exists today
public FooBuilder items(List<Item> items) {
    this.items = items;
    return this;
}

public FooBuilder item(Item item) {
     items.add(item);
     return this;
}

// or
public FooBuilder addItem(Item item) {
     items.add(item);
     return this;
}
```

Randgalt:
When `addSingleItemCollectionBuilders()` is enabled in options, collection types (`List`, `Set` and `Map`)
are handled specially. The setters for these types now create an internal collection and
items are added to that collection. Additionally, "adder" methods prefixed with
`singleItemBuilderPrefix()` are created to add single items to these collections.

The generated builder looks like this: https://gist.github.com/Randgalt/8aa487a847ea2acdd76d702f7cf17d6a

cc @tmichel

Closes #73

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -27,6 +27,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
+import static io.soabase.recordbuilder.processor.CollectionBuilderUtils.SingleItemsMetaDataMode.*;
 import static io.soabase.recordbuilder.processor.ElementUtils.getBuilderName;
 import static io.soabase.recordbuilder.processor.ElementUtils.getWithMethodName;
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generatedRecordBuilderAnnotation;
@@ -63,7 +64,7 @@ class InternalRecordBuilderProcessor {
         recordComponents = buildRecordComponents(record);
         uniqueVarName = getUniqueVarName();
         notNullPattern = Pattern.compile(metaData.interpretNotNullsPattern());
-        collectionBuilderUtils = new CollectionBuilderUtils(recordComponents, this.metaData.useImmutableCollections());
+        collectionBuilderUtils = new CollectionBuilderUtils(recordComponents, this.metaData);
 
         builder = TypeSpec.classBuilder(builderClassType.name())
                 .addAnnotation(generatedRecordBuilderAnnotation)
@@ -86,6 +87,8 @@ class InternalRecordBuilderProcessor {
             add1Field(component);
             add1SetterMethod(component);
             add1GetterMethod(component);
+            var collectionMetaData = collectionBuilderUtils.singleItemsMetaData(component, EXCLUDE_WILDCARD_TYPES);
+            collectionMetaData.ifPresent(meta -> add1CollectionBuilders(meta, component));
         });
         collectionBuilderUtils.addShims(builder);
         builderType = builder.build();
@@ -109,8 +112,7 @@ class InternalRecordBuilderProcessor {
                 builder.addModifiers(Modifier.PUBLIC);  // builders are top level classes - can only be public or package-private
             }
             // is package-private
-        }
-        else {
+        } else {
             builder.addModifiers(Modifier.PUBLIC);
         }
     }
@@ -234,6 +236,7 @@ class InternalRecordBuilderProcessor {
         codeBlockBuilder.add(";$]");
 
         var methodName = getWithMethodName(component, metaData.withClassMethodPrefix());
+        var singleItemsMetaData = collectionBuilderUtils.singleItemsMetaData(component, STANDARD_FOR_SETTER);
         var parameterSpecBuilder = ParameterSpec.builder(component.typeName(), component.name());
         addConstructorAnnotations(component, parameterSpecBuilder);
         var methodSpec = MethodSpec.methodBuilder(methodName)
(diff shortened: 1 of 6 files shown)


record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -631,6 +646,101 @@ private void addAccessorAnnotations(RecordClassType component, MethodSpec.Builde
...
+        }
+    }
+
+    private void add1MapBuilder(RecordClassType component) {
+        /*
+            For a single map record component, add a methods similar to:
+
+            public T addP(K key, V value) {
+                if (p == null) {
+                    p = new ArrayList<>();
```
tmichel, line 667: Shouldn't this be `p = new HashMap<>();`?

tmichel (commented):
Thanks for so quickly getting something up and running.

Am I correct that without the `useImmutableCollections()` option the allocated internal collections would be set in `build()`? 

I have two pet peeves:

1. Generating a varargs method incurs some performance cost because it allocates an array every time. In a hot loop this could be a problem. I never needed to add multiple items at once so I'm not sure that it is worth it to use varargs. I think calling `.addX(x1).addX(x2)` is not a big deal compared to `.addX(x1, x2)`. 
2. Special casing a single item collection for immutable copy can save an array allocation. `List.copyOf` internally calls `Collection.toArray` so using `List.of(list.get(0))` for a single item list can save that array allocation.

Randgalt:
> Am I correct that without the useImmutableCollections() option the allocated internal collections would be set in build()?

Yes - this keeps this change more consistent with the initial versions of the library which didn't do any post processing. If you want that additional processing set true `useImmutableCollections`.

> Generating a varargs method incurs some performance cost

Actually, I'd prefer not to have this as well. I only added it because the Immutables lib has it. I'll remove it.

> Special casing a single item collection for immutable copy can save an array allocation. List.copyOf internally calls...

Where is this?

tmichel:
> > Special casing a single item collection for immutable copy can save an array allocation. List.copyOf internally calls...
> 
> Where is this?

This happens in the `__list()`, `__set()`, etc helpers. I should have added that this observation is unrelated to the single adder feature. Sorry. 

This is from the linked gist:

```java
    }
```

This could be written as

```java
    if (o == null || o.isEmpty()) {
        return List.of();
    } else if (o.size() == 1) {
        return List.of(o.get(o));
    } else {
        return List.copyOf(o);
    }
}
```

I guess changing how immutable copies work would better be addressed in a separate PR so please ignore this. 😄

Randgalt:
> I guess changing how immutable copies work would better be addressed in a separate PR so please ignore this.

A PR would be appreciated!!

tmichel:
> > I guess changing how immutable copies work would better be addressed in a separate PR so please ignore this.
> 
> A PR would be appreciated!!

I'll try to take a stab at it tonight.

Randgalt:
@tmichel is this PR OK then?

tmichel:
I looked into changing the immutable copy methods and it does not worth it. `List.copyOf`, `Set.copyOf`, etc. are already quite optimal and copying does not really happen that often so there is no point further complicating the code. For the `Collections.unmodifiableList`, etc this might have made more sense but not with the new immutable collections.

Regarding this PR:

* Personally I prefer to have singular adder methods. So instead of `addItems(Item item)` I prefer `addItem(Item item)`. Getting the singular form of the plural is not trivial and probably needs a specialized library that can account for exceptions such as people -> person.
* Maybe use `Collection<T>` for adding elements to the internal collection for List and Set. `addAll` is coming from the `Collection` interface. 
* Now there is no way to clear the internal collection. I think it would make sense to either provide a `clearItems()` method or allow overwriting the entire contents of the collection. In the latter case `items(Collection<Item>)` would replace the contents of the internal collection. This would keep things consistent with the regular builder implementation. `addItems(Collection<Item>)` would add all items to the internal collection. This also necessitates the need for singular adder methods to avoid method name collisions and to clear things up.

Randgalt:
> * Personally I prefer to have singular adder methods. So instead of `addItems(Item item)` I prefer `addItem(Item item)`. Getting the singular form of the plural is not trivial and probably needs a specialized library that can account for exceptions such as people -> person.

While it makes sense grammatically, there's no good way to do this. The Immutables lib doesn't either. I'm -1 on this.

> * Maybe use `Collection<T>` for adding elements to the internal collection for List and Set. `addAll` is coming from the `Collection` interface.

+1

> * Now there is no way to clear the internal collection. I think it would make sense to either provide a `clearItems()` method or allow overwriting the entire contents of the collection. In the latter case `items(Collection<Item>)` would replace the contents of the internal collection. This would keep things consistent with the regular builder implementation. `addItems(Collection<Item>)` would add all items to the internal collection. This also necessitates the need for singular adder methods to avoid method name collisions and to clear things up.

+1

Randgalt:
@tmichel I've updated the code. In addition to the single item adders there's an adders take a `Stream` - that seems more useful than Collection or Iterator. Also, the existing builder method resets the internal collections. The gist has been updated: https://gist.github.com/Randgalt/8aa487a847ea2acdd76d702f7cf17d6a

tmichel:
>  In addition to the single item adders there's an adders take a `Stream` - that seems more useful than Collection or Iterator.

I disagree. `Stream` does make sense in some situations especially if you're mapping between layers but `Collection` is equally useful. I use a lot of builders in tests and in that context the need of passing a `Collection` to a builder method comes up quite often. I guess the beauty of method overload is that both can signatures are valid. `addItems(Stream<Item> stream)` and `addItems(Collection<Item> collection)`. 

Using the same name for adding a single item and adding all items from a collection/stream felt strange in the beginning but I guess I see the rationale behind it. Maybe the javadoc could highlight the difference.

> Also, the existing builder method resets the internal collections.

👍

Randgalt:
OK @tmichel - I added an additional version that takes an `Iterable` (which is more general than Collection and matches the Immutables lib).

tmichel:
This looks great now. I found a potential bug in the generated code:

```java
    public SingleItemsBuilder<T> strings(Collection<? extends String> strings) {
        this.strings = new ArrayList<>(strings);
        return this;
    }
```

Passing `null` will trip with a `NullPointerException` because `new ArrayList<>(...)` expects a non-null instance. I think it is useful to set `null` sometime to clear the collection. 

Proposed generated code: 

```java
    public SingleItemsBuilder<T> strings(Collection<? extends String> strings) {
        this.strings = strings == null ? null : new ArrayList<>(strings);
        return this;
    }
```

Randgalt:
>  I found a potential bug in the generated code:

Yeah - agree. I've updated the gist with the change. Good to merge now?

tmichel:
> Good to merge now?

Yes. Thanks for integrating my feedback into the PR.

PR #81, opened 2021-10-18
doc: update maven usage with default annotation processors discovery process

Stefanqn:
The current doc suggests a more complicated and possibly error-prone usage, which disables the default annotation processor discovery process, thus disabling (or at least specifying every) other processors e.g. Lombok.


Randgalt:
@Stefanqn please note the test failues.

Randgalt:
Thank you for the contribution. 🙇

PR #83, opened 2021-10-28
Optional functional methods for With

Randgalt:
When enabled, some functional methods are added to the `With` nested
class.

E.g.

```java
@RecordBuilder
record MyRecord<T>(String name, T value, int qty) implements MyRecordBuilder.With<T> {}

...

MyRecord<Thing> r = ...

var other = r.map((name, value, qty) -> new Other(...));

// also

r.accept((name, value, qty) -> {
   ...
});
```


ishanarora04:
@Randgalt I would like to contribute to this repo.  If you any have any feature/bug fixes on top of your mind , please let me know.

PR #88, opened 2022-02-11
Ignore generated files in jacoco coverage

### LINKED ISSUE #87: exclude from Jacoco coverage
Jacoco exclusion works by having` @Generated` annotation in bytecode, but the default `javax.annotation.processing.Generated` in only on source level.
https://github.com/jacoco/jacoco/blob/046ba52e9fd1288cf121fd1013c9019d71946954/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AnnotationGeneratedFilter.java

Can I add a custom annotation inside record-builder?

madisparn:
Fixes #87

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -95,6 +95,11 @@ public @interface RecordBuilder {
         String fromMethodName() default "from";
 
         /**
+         * The name to use for the from-to-wither inner class
+         */
+        String fromInnerClassName() default "WitherImpl";
+
+        /**
          * The name to use for the method that returns the record components as a stream
          */
         String componentsMethodName() default "stream";
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -31,6 +31,7 @@ import java.util.stream.Stream;
 import static io.soabase.recordbuilder.processor.CollectionBuilderUtils.SingleItemsMetaDataMode.*;
 import static io.soabase.recordbuilder.processor.ElementUtils.getBuilderName;
 import static io.soabase.recordbuilder.processor.ElementUtils.getWithMethodName;
+import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generatedCustomAnnotation;
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generatedRecordBuilderAnnotation;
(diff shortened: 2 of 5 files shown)


record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -571,35 +619,7 @@ public String p2() {
             });
             witherClassNameBuilder.add(">");
         }
-        var witherClassName = witherClassNameBuilder.build().toString();
-        var codeBuilder = CodeBlock.builder()
-                .add("return new $L", witherClassName)
```
madisparn, line 576: Jacoco does not check anonymous class annotations, so changing to `new @RecordBuilderGenerated With() {}` did not exclude the inner class from coverage

Randgalt:
Thanks for the submission - I'll review soon

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -94,6 +94,11 @@
          */
         String fromMethodName() default "from";
 
+        /**
+         * The name to use for the from-to-wither inner class
+         */
+        String fromInnerClassName() default "WitherImpl";
```
Randgalt, line 100: Can you explain this better? I'm -1 on this if its only purpose is to make Jacoco happy.

madisparn: It is not about making Jacoco happy, it is about not including generated code in test coverage.

Randgalt: Can you explain in detail please?

madisparn: image
I would like the record-builder generated classes to be excluded from coverage.

Randgalt: OK - but can't this be done by configuring Jacoco? Changing the code generation just so code coverage produces better reports seems backwards to me.

madisparn: To me it seems better if library users have to configure their build as little as possible. Excluding generated code from test coverage seems like it should be supported by annotation processor. Lombok does it.

Randgalt:
I see you closed this. I'll try to spend time on this to see if there's a good solution that meets both of our goals.

Randgalt:
I didn’t either. Can you reopen please?

madisparn:
Did not see a way to open this PR, created new one https://github.com/Randgalt/record-builder/pull/96

PR #92, opened 2022-03-04
Fix NPE for uninitialized non-null collection

### LINKED ISSUE #91: useImmutableCollections=true combined with @Nonnull on a record component result in NPE
### Discussed in https://github.com/Randgalt/record-builder/discussions/90

When `useImmutableCollections` is set to true, builder populates the record with empty collections rather than a null field. Generated code looks like:

```java
    public Removeme.R build() {
        return new Removeme.R(__list(l));
    }

    }
```

So far, so good. Now, let's add @Nonnull annotation on the record component. Building the record will result in a NullPointerException because builder check for nullity *before* `__list()` is invoked.

```java
    public Removeme.R build() {
        Objects.requireNonNull(l, "l is required");
        return new Removeme.R(__list(l));
    }
```

Declaring a collection to not be null is likely the most common case (who want to deal with both nullity and emptiness?). 

I'm not familiar with record-builder codebase but will submit a tentative PR soon.

cykl:
Closes #91

This is a tentative patch for #91. I'm not familiar with record-builder code base, so I prefer to share an early draft and get feedback before trying to polish it. 

The idea is to move  collection shims invocation before null checks. 

Before:

```
    public Removeme.R build() {
        Objects.requireNonNull(l, "l is required");
        return new Removeme.R(__list(l));
    }
```

After:
```
    public Removeme.R build() {
        l = __list(l);
        Objects.requireNonNull(l, "l is required");
        return new Removeme.R(l);
    }
```

Shortcomings:
 * Break backward compatibility. I find that current behavior is a bug, but others might disagree and rely on current semantic. See updated test for impact.
 * Current implementation will emit many useless `x = x;` assignments. If that's a problem we can filter non collection components.

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -521,6 +521,14 @@ class InternalRecordBuilderProcessor {
         */
 
         var codeBuilder = CodeBlock.builder();
+
+        IntStream.range(0, recordComponents.size()).forEach(index -> {
+            var recordComponent = recordComponents.get(index);
+            codeBuilder.add("$[$L = ", recordComponent.name());
+            collectionBuilderUtils.add(codeBuilder, recordComponents.get(index));
+            codeBuilder.add(";\n$]");
+        });
+
         addNullCheckCodeBlock(codeBuilder);
         codeBuilder.add("$[return ");
         if (metaData.useValidationApi()) {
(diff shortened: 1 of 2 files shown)


Randgalt:
@cykl thanks for the PR. I'm thinking that the `Objects.requireNonNull` calls are not needed actually. The values will never be null right? Wouldn't a simpler solution be to not emit the `Objects.requireNonNull` calls when `useImmutableCollections` is `true`?

cykl:
If think that your are right. `Objects.requireNonNull` is not required for collection members when `useImmutableCollections` is `true` since the shim ensure that it is not. However, it is still required for non collection members.

I still have to check what happen if caller pass null to the builder (I expect the "setter" to catch that but haven't checked).

Because I'm not yet comfortable with all the features and code of record-builder, my approach was to write the safest & dumbest code. It should compose nicely with any other feature at the expense of prettiness of generated code. I'm not too worried about performances as I think JIT will eliminate all those useless checks and assignments.

If we agree on the goal and you think there are no obvious pitfalls to avoid, I can try to improve the patch.

Randgalt:
With `useImmutableCollections ` passing nulls to the builders is fine. It will get converted to an empty collection.

Randgalt:
Closing in favor of #98

PR #94, opened 2022-03-07
Add flag to add concrete setters for optionals

mads-b:
In my adventure to replace Immutables.org, I stumbled upon yet another nifty feature:

If you are setting an optional field, you are probably intending to set a concrete value.

So here is this PR that adds setters that take concrete values and wrap them in optional prior to insertion into the record.

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/OptionalType.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/OptionalType.java
@@ -39,10 +39,9 @@ public record OptionalType(TypeName typeName, TypeName valueType) {
 
     static Optional<OptionalType> fromClassType(final ClassType component) {
         if (isOptional(component)) {
-            if (!(component.typeName() instanceof ParameterizedTypeName)) {
+            if (!(component.typeName() instanceof ParameterizedTypeName parameterizedType)) {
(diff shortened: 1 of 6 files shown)


Randgalt:
For completeness, please add a test Record that has `Optional` components but with `addConcreteSettersForOptional` not enabled. Basically, make sure all the permutations are tested.

Randgalt:
@mads-b Looks good - please rebase with master

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/OptionalType.java
```diff
@@ -0,0 +1,62 @@
...
+        if (component.typeName().equals(optionalType)) {
+            return true;
+        }
+        return (component.typeName() instanceof ParameterizedTypeName parameterizedTypeName)
+                && parameterizedTypeName.rawType.equals(optionalType);
+    }
+
+    static Optional<OptionalType> fromClassType(final ClassType component) {
+        if (isOptional(component)) {
+            if (!(component.typeName() instanceof ParameterizedTypeName)) {
```
Randgalt, line 42: Can be:

```
if (!(component.typeName() instanceof ParameterizedTypeName parameterizedType))
```

PR #95, opened 2022-03-07
Make withers and getters optional features

mads-b:
My generated builders are getting rather large for huge records and it is an unnecessary strain on javac. Additionally, this can be a problem for e.g. android users that want to limit their app sizes and avoid the upper limits of class count in DEX.

I was intending on making the enableGetters=false flag also remove the getters on the wither, but that turned out to be a bit of an adventure in rewrite, so I omitted that particular part just to keep the PR small.


Randgalt:
Hi - can you please add an instance of a Record that has these features enabled? It will serve as a test of the feature.

mads-b:
Done! Rather minimal example, but at least it ensures the code gets run

PR #98, opened 2022-03-20
Fix NPE for uninitialized non-null collection

### LINKED ISSUE #91: useImmutableCollections=true combined with @Nonnull on a record component result in NPE

Randgalt:
Closes #91

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
         <maven-shade-plugin-version>3.2.1</maven-shade-plugin-version>
         <maven-release-plugin-version>2.5.3</maven-release-plugin-version>
         <maven-jar-plugin-version>3.2.0</maven-jar-plugin-version>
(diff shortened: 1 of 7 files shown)


cykl:
Thanks @Randgalt!

PR #99, opened 2022-03-21
Don't include `@Valid` on base interfaces

### LINKED ISSUE #97: Withers ConstraintDeclarationException
When I want to validate a record as a field of another record which implements `With` interface, the `javax.validation.ConstraintDeclarationException` is thrown.

Repro:

```java
@RecordBuilder
public record Request(

  @NotNull
  @Valid
  Part part)

  implements RequestBuilder.With {

  public record Part(@NotBlank String name) {}

}
```

Error message:

```java
HV000131: A method return value must not be marked for cascaded validation more than once in a class hierarchy, but the following two methods are marked as such: Request#part(), With#part().
```

It works without `With`.

Randgalt:
The validation API doesn't accept `@Valid` on base interfaces. Filter
them out.

Fixes #97


Randgalt:
@nikolavojicic please review

nikolavojicic:
Yes, that fixes Issue 97.

PR #102, opened 2022-04-03
Generated equals() and the default with() method were missing generics

### LINKED ISSUE #101: Generic records produce code with `rawtypes` and `unchecked` warnings
At work, we are making great use of your library. Thanks!

That said, we introduced our first record using generics the other day, and found the generated Builder class to suffer from a few javac warnings, namely `unchecked` and `rawtypes`. We have a very low tolerance for compiler warnings, so this blocks us from using @RecordBuilder for these types.

We use JDK 17

Minimal example:
```java
@RecordBuilder
public record Foo<T>(List<T> items) {}
```

Problematic generated code snippets:
```java
// FooBuilder

@Override
public boolean equals(Object o) {
    return (this == o) || ((o instanceof FooBuilder r)    // <- javac flags [rawtypes] warning here
            && Objects.equals(items, r.items));
}

// ...

default FooBuilder<T> with() {
    return new FooBuilder(items());  // <- javac flags [rawtypes] and [unchecked] warnings here
}
```

It would be ideal if `equals()` could appease the compiler with `(o instanceof FooBuilder<?> r)` and `with()` could similarly with `new FooBuilder<>(items())`.

Barring that, perhaps a `@SuppressWarnings({"unchecked","rawtypes"})` on these methods or on the class.

Randgalt:
Generated equals() and the default with() method were missing generic angle brackets causing compiler warnings

Closes #101

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -232,7 +232,7 @@ class InternalRecordBuilderProcessor {
             }
          */
         var codeBlockBuilder = CodeBlock.builder()
-                .add("return new $L(", builderClassType.name());
+                .add("return new $L$L(", builderClassType.name(), typeVariables.isEmpty() ? "" : "<>");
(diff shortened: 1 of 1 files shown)


Randgalt:
@daveschoutens please review

PR #104, opened 2022-04-07
Options/changes so that Jacoco checks don't fail

### LINKED ISSUE #87: exclude from Jacoco coverage

Randgalt:
- Added new optional Annotation `@RecordBuilderGenerated` - Jacoco ignores
classes with any annotation names "\*Generated\*" but it needs to be class retained.
For backward compatibility this annotation is not added by default (though it's been
added to `@RecordBuilderFull`). There is a new option to enable it.
- The from with method now uses an internal static class instead of an anonymous
inner class so that the annotation can be on this as well. This new class's name
is configurable in the options.

Thanks to user @madisparn for initial PR and issue report.

Fixes #87

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -33,6 +33,7 @@ import static io.soabase.recordbuilder.processor.CollectionBuilderUtils.SingleIt
 import static io.soabase.recordbuilder.processor.ElementUtils.getBuilderName;
 import static io.soabase.recordbuilder.processor.ElementUtils.getWithMethodName;
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generatedRecordBuilderAnnotation;
+import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.recordBuilderGeneratedAnnotation;
 
 class InternalRecordBuilderProcessor {
     private final RecordBuilder.Options metaData;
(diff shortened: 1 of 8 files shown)


record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -560,63 +567,83 @@ private CodeBlock buildCodeBlock() {
...
-                    public String p1() {
-                        return from.p1();
-                    }
-
-                    @Override
-                    public String p2() {
-                        return from.p2();
-                    }
-                };
+            private static final class _FromWith implements MyRecordBuilder.With {
```
madisparn, line 583: Class name will be `MyRecordBuilder_FromWith` ?

madisparn:
Looks good

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
Randgalt: `MyRecordBuilder._FromWith` - The `_` is commonly used to hide internal names. But, I'm open to other names.

madisparn: `_FromWith` is good, I just did not read the code properly

Issue #114, opened 2022-05-20
Heavy copying of collections


freelon:
If using the `useImmutableCollections` and `addSingleItemCollectionBuilders` options the generated builder does copy collections on every change on the record. Example:

    @RecordBuilderFull
    public record Example(String name, List<Foo> foo) implements ExampleBuilder.With {}
    ...
    Example e1 = ExampleBuilder.builder().name("a").foo( *some large collection*).build();
    Example e2 = e1.with().name("b").build();
    Example e3 = e2.with().name("c").build();

Whenever I do something like `example.with().name("...").build()` internally `with()` creates a new ExampleBuilder which instantiates a new ArrayList with all items from my old example object. Then calling `build()` creates a new immutable list of the array list. If I have a large collection and change other fields of my record frequently a lot (2 per record with-...-build-cycle) of list-copying is done in the background, although the list is never changed. So in the above example my large collection would already been copied six times without need.

I would prefer if the list was only copied into a new ArrayList if it actually changes. I think this can be done rather easily by using for instance a flag for every collection to indicate if the collection was modified. If a change is to be made and the flag is false, copy the current collection into a new ArrayList and then set the flag. After that, just add the collection item(s).

I'd be glad to create a PR if you're interested. Thanks for the great library so far :)

Randgalt:
Hi - yes a PR would be appreciated.

Randgalt:
Hi @freelon - I found a bug with the PR and reverted it. The problem is the determination if a collection/map has been changed to mutable is a check that it's an `ArrayList` or a `HashMap`, etc. This breaks if the source record's list, map is passed in. You can see this with this test:

```
    @Test
    void testSourceRecordNotModified() {
        List<Number> list = new ArrayList<>();
        Map<Number, FullRecord> map = new HashMap<>();
        FullRecord record = new FullRecord(list, map, "a");
        FullRecordBuilder builder = FullRecordBuilder.builder(record);
        builder.addNumbers(10);
        builder.addFullRecords(10, new FullRecord(List.of(), Map.of(), "b"));
        Assertions.assertTrue(list.isEmpty());
        Assertions.assertTrue(map.isEmpty());
    }
```

The solution is to add a `boolean` flag for each collection to test if it's been converted to modifiable.

PR #116, opened 2022-05-22
Collection copying only when changed

freelon:
For #114 

Currently the test `TestCollections.testRecordBuilderOptionsCopied` fails, since I had to change the signature of the `__list()` shim methods. The problem is, that the code as on master generates setters for collections with different signatures: 
If `addSingleItemCollectionBuilders == true`: `someList(Collection<? extends ListItem> someList)`, but otherwise the `someList` parameter is of type `List<...>`. I didn't want to change that behavior together with the other ticket, though (see #117 ).

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
@@ -31,15 +31,30 @@ class CollectionBuilderUtils {
     private final String setShimName;
     private final String collectionShimName;
 
+    private final String listMakerMethodName;
+    private final String mapMakerMethodName;
+    private final String setMakerMethodName;
+
     private boolean needsListShim;
     private boolean needsMapShim;
     private boolean needsSetShim;
     private boolean needsCollectionShim;
 
+    private boolean needsListMutableMaker;
+    private boolean needsMapMutableMaker;
+    private boolean needsSetMutableMaker;
+
     private static final TypeName listType = TypeName.get(List.class);
     private static final TypeName mapType = TypeName.get(Map.class);
     private static final TypeName setType = TypeName.get(Set.class);
     private static final TypeName collectionType = TypeName.get(Collection.class);
+
+    private static final Class<?> mutableListType = ArrayList.class;
+    private static final Class<?> mutableMapType = HashMap.class;
+    private static final Class<?> mutableSetType = HashSet.class;
+    private static final TypeName mutableListTypeName = TypeName.get(mutableListType);
+    private static final TypeName mutableMapTypeName = TypeName.get(mutableMapType);
+    private static final TypeName mutableSetTypeName = TypeName.get(mutableSetType);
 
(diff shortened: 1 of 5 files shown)


Randgalt:
@freelon there are test failures

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
```diff
@@ -187,34 +255,52 @@ private boolean hasWildcardTypeArguments(ParameterizedTypeName parameterizedType
...
+                .addParameter(extendedParameterizedType, "o")
                 .addStatement(code)
                 .build();
     }
 
-    private MethodSpec buildCollectionsMethod() {
+    private MethodSpec buildMutableMakerMethod(String name, TypeName mutableCollectionType, ParameterizedTypeName parameterizedType, TypeVariableName... typeVariables) {
+        var nullCase = CodeBlock.of("if (o == null) return new $T<>()", mutableCollectionType);
+        var isMutableCase = CodeBlock.of("if (o instanceof $T) return o", mutableCollectionType);
+        var defaultCase = CodeBlock.of("return new $T(o)", mutableCollectionType);
```
Randgalt, line 283: Should be:

```
var defaultCase = CodeBlock.of("return new $T<>(o)", mutableCollectionType);
```

Randgalt:
> Currently the test TestCollections.testRecordBuilderOptionsCopied fails, since I had to change the signature of the __list() shim methods. The problem is, that the code as on master generates setters for collections with different signatures:

Instead of making a backward incompatible change can't we add a 2 `__list` shims? One for `List` and one for `Collection`?

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
freelon: :heavy_check_mark:

freelon:
> Instead of making a backward incompatible change can't we add a 2 __list shims? One for List and one for Collection?

Of course, but why keep `__list(List<X> o)`? It's more restrictive and does the same as `__list(Collection ...)`. Also, it's a private method, so I'd think backwards compatibility is not much of an issue.

Randgalt:
> > Currently the test TestCollections.testRecordBuilderOptionsCopied fails, since I had to change the signature of the __list() shim methods. The problem is, that the code as on master generates setters for collections with different signatures:
> 
> Instead of making a backward incompatible change can't we add a 2 `__list` shims? One for `List` and one for `Collection`?

True - but in that case the test can be fixed here right?

freelon:
Yes, I already did that. I just didn't want to lead with changing tests ;)

freelon:
Is there anything more to do on this or can it be merged? From my POV all the discussed points are solved.

Randgalt:
I apologize - I got bogged down with work. I'll get to this soon (hopefully this week)

Randgalt:
LGTM - thank you very much

PR #130, opened 2022-10-19
Fix addSingleItemCollectionBuilders when useImmutableCollections is false

### LINKED ISSUE #129: aa027af causes compile errors
Hello, The aforementioned commit is preventing me from upgrading from v33 to v34. The issue:

I have a quite simple record which should create a builder:
```
public record CombinedFields(
    @JsonValue String combinedField,
    List<CombinedFieldsLine> lines) implements CombinedFieldsBuilder.Bean {}
    ```
But now, the generated builder contains a setter with an undefined method (shim?)

```
    /**
     * Re-create the internally allocated {@code List<CombinedFields.CombinedFieldsLine>} for {@code lines} by copying the argument
     */
    public CombinedFieldsBuilder setLines(
            Collection<? extends CombinedFields.CombinedFieldsLine> lines) {
        this.lines = __list(lines);
        return this;
    }
    ```
It's this __list method that does not exist. Is this a bug or something I need to change in my configuration?

Randgalt:
- Removed `SingleItemsMetaDataMode.STANDARD` as it wasn't used
- Removed `needs*MutableMaker` - always add them when corresponding `needs*Shim` is true
- Merged `addMutableMakers()` into `addShims()`
- Always use mutable makers when addSingleItemCollectionBuilders is true

Fixes #129

cc @mads-b, @freelon, @tmichel

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
@@ -42,10 +42,6 @@ class CollectionBuilderUtils {
     private boolean needsSetShim;
     private boolean needsCollectionShim;
 
-    private boolean needsListMutableMaker;
-    private boolean needsMapMutableMaker;
-    private boolean needsSetMutableMaker;
-
     private static final Class<?> listType = List.class;
     private static final Class<?> mapType = Map.class;
     private static final Class<?> setType = Set.class;
@@ -93,7 +89,6 @@ class CollectionBuilderUtils {
     }
 
     enum SingleItemsMetaDataMode {
-        STANDARD,
         STANDARD_FOR_SETTER,
         EXCLUDE_WILDCARD_TYPES
     }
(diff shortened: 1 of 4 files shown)


freelon:
- `needs*Shim` is now used for shim and mutable makers. It should be renamed to reflect its meaning.
- if only `addSingleItemCollectionBuilders`is true (i.e. lists are mutable) the setters for collections always create an immutable copy of the collection passed to it
  - if an item is added to that list another copy is made to make it mutable again
  - not sure what's the way to go here, but I'd prefer to be still able to set a fixed list instance (see [this discussion](https://github.com/Randgalt/record-builder/issues/129#issuecomment-1272293804) for a more explicit way of creating a new list with all items of a collection (clear and addAll))

freelon:
On a separate note, now if I use the collection setter (addSingleItemCollectionBuilders=true) with e.g. an ArrayList I'll get an exception later if I try `myRecords.someList().add(...)` because it became immutable without me asking for immutable lists..

Randgalt:
I see - these two features may be hopelessly broken: `addSingleItemCollectionBuilders` is the real troublesome one. I'll think more about this and possibly do something similar to [this](https://github.com/Randgalt/record-builder/issues/129#issuecomment-1272293804

PR #133, opened 2022-12-04
use configured build method name instead of hard-coded name

sebhoss:
This fixes an issue when a custom `build` method is configured and nested records are used together with the wither feature enabled.


Randgalt:
Thanks for the PR

PR #134, opened 2022-12-06
feat: add optional suffix to static builder

RichardTree:
Not sure what you think of this, my team are coming from using Immutables heavily and are fond of their FieldName.of() syntax convention for static builders.

I’m assuming my change would enable us to call a static builder like FieldNameBuilder.FieldNameOf().

When statically imported, would read succinctly as _FieldNameOf()_

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -363,7 +363,7 @@ class InternalRecordBuilderProcessor {
             }
          */
         CodeBlock codeBlock = buildCodeBlock();
-        var builder = MethodSpec.methodBuilder(recordClassType.name())
+        var builder = MethodSpec.methodBuilder(recordClassType.name() + metaData.staticBuilderSuffix())
                 .addJavadoc("Static constructor/builder. Can be used instead of new $L(...)\n", recordClassType.name())
                 .addTypeVariables(typeVariables)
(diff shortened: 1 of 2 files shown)


Randgalt:
Thanks for the PR - I'll look at this soon.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -363,7 +363,7 @@ public static MyRecord(int p1, T p2, ...) {
             }
          */
         CodeBlock codeBlock = buildCodeBlock();
-        var builder = MethodSpec.methodBuilder(recordClassType.name())
+        var builder = MethodSpec.methodBuilder(recordClassType.name() + metaData.staticBuilderSuffix())
```
codefish1, line 366:
```suggestion
           var staticBuilderName = Optional.of(metaData.staticBuilderSuffix()).filter(e -> !e.isEmpty()).map(e -> lowerCaseFirstLetter(recordClassType.name())+ e).orElse(recordClassType.name());
        var builder = MethodSpec.methodBuilder(staticBuilderName)
```

```
   private String lowerCaseFirstLetter(String value) {
        return value.substring(0,1).toLowerCase() + value.substring(1);
    }
```


Might be nicer to lowercase the first letter of the static builder as well so instead of `FirstnameOf` you have `firstnameOf` for example

Randgalt:
My initial reaction is that RecordBuilder is getting too many options. We've already had some incompatible ones. We may need to re-think the customization scheme.

PR #135, opened 2023-01-01
Configurable modifiers on static builders

codefish1:
### The use case
Ensure people use the static builder and don't new up the record directly by restricting access to the records constructor.  This becomes more important when 50+ developers are working on the same code base, by using the compiler to help enforce the stylist approaches.

### Why Do we need this?

Currently with records we can't make the constructor private, it will have the same visibility as the record it's self.  In addition when using this library, the static builder will also have the same visibility as the record.

### What does this PR add?

This PR enables you to configure the modifiers on the static builder such that you can set the static builder to have a public modifier whereas the record is package-protect, so when your records are in a domain package they can't be accessed directly and new'ed up.

### Discussion points
 1. Do we want to make this configurable like in this PR or a simple boolean flag

### Outstanding work

Assuming you are happy with the principle and approach, I need to update the documentation.

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -15,6 +15,7 @@
  */
 package io.soabase.recordbuilder.core;
 
+import javax.lang.model.element.Modifier;
 import java.lang.annotation.*;
 
 @Retention(RetentionPolicy.SOURCE)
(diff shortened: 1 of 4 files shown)


Randgalt:
This is good - I like the option as is. Please update the doc.

codefish1:
Is there a page which describes the various options?  All I can see is a link to the RecordBuilder annotation to view the javadoc in there

Randgalt:
> Is there a page which describes the various options? All I can see is a link to the RecordBuilder annotation to view the javadoc in there

I guess not.

Randgalt:
Thanks for the submission

Issue #122, opened 2022-07-19
Builder defaults for collections are inconsistent with output record


fprochazka:
Hi,
awesome library! Love it! I just have a small suggestion...

If I have a record with any collection

```java
@MyRecordBuilder // with useImmutableCollections = true, interpretNotNulls = true
public record Costs(
    @Nullable BigDecimal shipping,
    Map<String, Object> extraData
) { }
```

* the default values when the builder initializes for all fields is `null`
* calling `CostsBuilder.builder().build().extraData()` builds instance of `Costs` that has `extraData` initialized with empty `Map.of()` and does not return `null`
    * which means `CostsBuilder.builder().build().extraData().get("value")` **will not** throw NPE
* calling `CostsBuilder.builder().extraData()` returns the default `null`
    * which means `CostsBuilder.builder().extraData().get("value")` **will** throw NPE

I'm proposing that since you're adding special handling of collections and always initializing them, they should also be initialized to empty collections when the empty builder is created, so that the second case will also not throw an NPE

Randgalt:
Could these be combined into a single PR? Maybe with a new option to avoid breaking existing code.

fprochazka:
@Randgalt I've created two issues because I expected others to have a different opinion and you might be willing to accept only one part of the proposal, but both of the issues are related in my mind.

hofiisek:
Isn't this one solved by [this](https://github.com/Randgalt/record-builder/pull/138)?

PR #138, opened 2023-01-14
issue-122: updated add1GetterMethod logic in useImmutableCollections=true case

epm-dev-priporov:
fix for #122 
I am not sure that this is something exactly you would like to expect, but I reproduced and tested this case, seems it works. Would be glad to get any feedback. 
Please let me know if everything is fine - I would like to take issue 123 in that case as well

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -938,9 +938,21 @@ class InternalRecordBuilderProcessor {
                 .addModifiers(Modifier.PUBLIC)
                 .addAnnotation(generatedRecordBuilderAnnotation)
                 .returns(component.typeName())
-                .addStatement("return $L", component.name());
+                .addCode(build1codeBlock(component));
         addAccessorAnnotations(component, methodSpecBuilder, __ -> true);
         builder.addMethod(methodSpecBuilder.build());
+    }
+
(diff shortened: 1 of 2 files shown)


Randgalt:
> I reproduced and tested this case

Can you add a test for this please?

epm-dev-priporov:
yes, I will add a bit late

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -938,11 +938,23 @@ public T p() {
                 .addModifiers(Modifier.PUBLIC)
                 .addAnnotation(generatedRecordBuilderAnnotation)
                 .returns(component.typeName())
-                .addStatement("return $L", component.name());
+                .addCode(build1codeBlock(component));
         addAccessorAnnotations(component, methodSpecBuilder, __ -> true);
         builder.addMethod(methodSpecBuilder.build());
     }
 
+    private CodeBlock build1codeBlock(RecordClassType component) {
```
Randgalt, line 946: Please rename `checkReturnShim()` or something similar

Randgalt (approved):

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
epm-dev-priporov: done

Randgalt:
Thank you for the contribution!

PR #144, opened 2023-02-28
Fix javadoc builder core

varunu28:
**Changes in the PR**
  - Fixed javadoc to specify the return type as `collection` instead of `list` which is a specific data structure. 
  
Screenshot 2023-02-28 at 6 53 43 AM
 - Removed blank lines from javadoc and replaced with `<br>` tag.Reference: [How to add line break in Javadoc](https://forums.oracle.com/ords/apexds/post/easy-javadoc-question-how-to-make-newline-1928)
 - Some rewording in javadoc which IDE complaints about and replaced with auto-suggestion.

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -1,9 +1,12 @@
 /**
  * Copyright 2019 Jordan Zimmerman
+ * <br>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
+ * <br>
  * http://www.apache.org/licenses/LICENSE-2.0
+ * <br>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2019 Jordan Zimmerman
- * <br>
+ *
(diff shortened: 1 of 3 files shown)


record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -1,12 +1,12 @@
 /**
  * Copyright 2019 Jordan Zimmerman
- *
+ * <br>
```
Randgalt, line 3: This is coming from the license file and is not hand coded. This is the standard ALS license. I think this is not correct.

varunu28: Makes sense. I have reverted the line break change. Though the IDE still suggests adding a line break tag or a paragraph tag. Otherwise adding the line break doesn't makes any affect as it won't be rendered as a line break in Javadoc.

tisonkun: To avoid such case, I open https://github.com/Randgalt/record-builder/pull/145 and you can give it a review :)

varunu28:
@Randgalt Apologies for misunderstanding. As there was no response since last discussion, I felt either this is no longer required or has been missed in the review. 

Sorry for abrupt reaction on my end.

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -27,24 +24,24 @@
...
     @interface Include {
         /**
-         * @return list of classes to include
+         * @return collection of classes to include
          */
         Class<?>[] value() default {};
 
         /**
-         * Synonym for {@code value()}. When using the other attributes it maybe more clear to
+         * Synonym for {@code value()}. When using the other attributes it maybe clearer to
```
Randgalt, line 35: This should now be `may be clearer`

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordInterface.java
```diff
@@ -1,12 +1,9 @@
 /**
```
Randgalt: Please rebase with master as it has the other PR that fixes the license plugin

PR #147, opened 2023-03-12
Delete any pre-existing classfile for RecordInterface

### LINKED ISSUE #139: Cannot invoke "javax.lang.model.element.ExecutableElement.getAnnotationMirrors()" because the return value of "javax.lang.model.element.RecordComponentElement.getAccessor()" is null
We use record-builder on a project and we often get compilation errors with the following message:
`
Cannot invoke "javax.lang.model.element.ExecutableElement.getAnnotationMirrors()" because the return value of "javax.lang.model.element.RecordComponentElement.getAccessor()" is null
`

The errors happen either in Intellij IDEA or via the CLI with Maven. Running a re-build in Intellij or running the command again fixes the issue (usually a `mvn verify`).

I suspect it could come from the two tools using a different JDK or compilation options and each generating slightly different bytecode. However I've made sure that the same JDK (17.0.6) is used in both cases.

I've managed to capture a stack trace below. I'm not confident about being able to provide a reproducer project but if the error message is not helpful in understanding the problem I can give it a shot.

```
Caused by: java.lang.NullPointerException: Cannot invoke "javax.lang.model.element.ExecutableElement.getAnnotationMirrors()" because the return value of "javax.lang.model.element.RecordComponentElement.getAccessor()" is null
    at io.soabase.recordbuilder.processor.InternalRecordBuilderProcessor.lambda$buildRecordComponents$2 (InternalRecordBuilderProcessor.java:143)
    at java.util.stream.ReferencePipeline$3$1.accept (ReferencePipeline.java:197)
    at java.util.Iterator.forEachRemaining (Iterator.java:133)
    at java.util.Spliterators$IteratorSpliterator.forEachRemaining (Spliterators.java:1845)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:509)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:499)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential (ReduceOps.java:921)
    at java.util.stream.AbstractPipeline.evaluate (AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect (ReferencePipeline.java:682)
    at io.soabase.recordbuilder.processor.InternalRecordBuilderProcessor.buildRecordComponents (InternalRecordBuilderProcessor.java:143)
    at io.soabase.recordbuilder.processor.InternalRecordBuilderProcessor.<init> (InternalRecordBuilderProcessor.java:66)
    at io.soabase.recordbuilder.processor.RecordBuilderProcessor.processRecordBuilder (RecordBuilderProcessor.java:168)
    at io.soabase.recordbuilder.processor.RecordBuilderProcessor.process (RecordBuilderProcessor.java:76)
    at io.soabase.recordbuilder.processor.RecordBuilderProcessor.lambda$process$0 (RecordBuilderProcessor.java:54)
    at java.lang.Iterable.forEach (Iterable.java:75)
    at io.soabase.recordbuilder.processor.RecordBuilderProcessor.lambda$process$1 (RecordBuilderProcessor.java:54)
    at java.lang.Iterable.forEach (Iterable.java:75)
    at io.soabase.recordbuilder.processor.RecordBuilderProcessor.process (RecordBuilderProcessor.java:54)
    at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor (JavacProcessingEnvironment.java:1023)
    ....
```

Randgalt:
Fixes #139

It seems the existence of an old .class file for the record being generated by `@RecordInterface` is exposing a bug in javac. This change checks for the existence of the class file and deletes. It's an ugly hack but seems to work.

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/RecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/RecordBuilderProcessor.java
@@ -31,7 +31,10 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
 import javax.tools.Diagnostic;
+import javax.tools.FileObject;
 import javax.tools.JavaFileObject;
+import javax.tools.StandardLocation;
+import java.io.File;
 import java.io.IOException;
 import java.io.Writer;
 import java.util.Optional;
@@ -193,6 +196,13 @@ public class RecordBuilderProcessor
 
         Filer filer = processingEnv.getFiler();
         try {
+            FileObject existingClassFile = filer.getResource(StandardLocation.CLASS_OUTPUT, packageName, classType.name() + ".class");
+            String path = existingClassFile.toUri().getPath();
+            boolean delete = new File(path).delete();
+            if (delete) {
+                processingEnv.getMessager().printMessage(Diagnostic.Kind.NOTE, "Deleted pre-existing record class at " + path + " " + delete);
+            }
+
             String fullyQualifiedName = packageName.isEmpty() ? classType.name() : (packageName + "." + classType.name());
             JavaFileObject sourceFile = filer.createSourceFile(fullyQualifiedName);
             try (Writer writer = sourceFile.openWriter()) {


Randgalt:
@youribonnaffe and @cykl

As mentioned in #139 I've been able to reproduce the problem with IntelliJ when I manually compile a single interface annotated with `@RecordInterface`. In this instance, the presence of a .class file of the generated record from a previous compile is the source of the problem. If that .class file is deleted then compilation works correctly. This PR adds a hack that deletes any .class file for the record. Please let me know if it fixes the issue for you as well. If it does, we can figure out a way to make it less of a hack.

Issue #123, opened 2022-07-19
Allow nullable collections with useImmutableCollections=true


fprochazka:
Hi,
somewhat related to #122, I'd like to propose **allowing** collections to become nullable, because right now `useImmutableCollections = true` does two things

* it adds the conversion to immutable collections (awesome!)
* but also enforces that the output record can never have the collection values nullable

I'm proposing to change the behaviour based on `interpretNotNulls` value.

* `useImmutableCollections = true && interpretNotNulls = false`
    * mapping will be `return (o != null) ? Map.copyOf(o) : null;`
* `useImmutableCollections = true && interpretNotNulls = true`
    * and field is determined to be nullable  
        * mapping will be `return (o != null) ? Map.copyOf(o) : null;`
    * and field is determined to be notnull
        * mapping will be `return (o != null) ? Map.copyOf(o) : Map.of();`
* `useImmutableCollections = true`
    * current behaviour 

This should also (consistently) affect the default value for collections mentioned in #122

hofiisek:
I started working on this and so far I got to a point where record components return `null` when a new `useNullableCollections` option is set to `true` (default is false for backwards compatibility). 

I decided to create a new option for it to not break any existing code as I'd not be surprised if library users were relying on these collection/list/set/map components always being non-null. Please let me know if it should be done otherwise.

PR #156, opened 2023-10-31
Issue #123: support nullable collections

hofiisek:
Solves https://github.com/Randgalt/record-builder/issues/123

There's a new `allowNullableCollections` option that does the following:
- if set to `false` (default value for BC), does nothing
- if set to `true` and `interpretNotNulls = false`, a list/set/map/collection record component will return null if value is not set
- if set to `true` and `interpretNotNulls = true`, a list/set/map/collection record component will return null only if it's determined to be nullable, i.e. isn't annotated by any of the not-null patterns

All branches mentioned above should be covered by the `TestNullableCollectionsBuilder` test

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
@@ -26,7 +26,6 @@ import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generate
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.recordBuilderGeneratedAnnotation;
 
 class CollectionBuilderUtils {
-
     private final boolean useImmutableCollections;
     private final boolean useUnmodifiableCollections;
     private final boolean allowNullableCollections;
@@ -410,7 +409,7 @@ class CollectionBuilderUtils {
                     collectionsTypeName, kType, vType, parameterizedType, collectionsTypeName, kType, vType);
         }
 
-        throw new IllegalStateException("Cannot build shim method for" + mainType);
+        throw new IllegalStateException("Cannot build shim method for " + mainType);
     }
 
     private MethodSpec buildNullableShimMethod(String name, TypeName mainType, Class<?> abstractType,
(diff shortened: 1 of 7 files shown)


Randgalt:
Thanks for this. I'll have a look very soon

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
```diff
@@ -20,20 +20,32 @@
 
 import javax.lang.model.element.Modifier;
 import java.util.*;
+import java.util.regex.Pattern;
 
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.generatedRecordBuilderAnnotation;
 import static io.soabase.recordbuilder.processor.RecordBuilderProcessor.recordBuilderGeneratedAnnotation;
 
 class CollectionBuilderUtils {
+
```
Randgalt, line 29: nit: blank line

Randgalt:
I tried a class that only has `@RecordBuilder.Options(allowNullableCollections = true)` and it's a NOP regarding specialized collections. Should we emit a warning or error? Or, should `allowNullableCollections = true` always imply enabling `useImmutableCollections`?

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/CollectionBuilderUtils.java
hofiisek: fixed

hofiisek:
> Thanks for this PR - it looks very good.
> 
> I tried a class that only has `@RecordBuilder.Options(allowNullableCollections = true)` and it's a NOP regarding specialized collections. Should we emit a warning or error? Or, should `allowNullableCollections = true` always imply enabling `useImmutableCollections`?

I see, didn't think of that combination before. Anyway as far as I understand: if only `allowNullableCollections = true` is set, then `useUnmodifiableCollections` as well `useImmutableCollections` is `false`, and then the record components (and builder getters) should just return whatever-value you set to them, right? Without any shim calls etc. Then I'd say that NOP is actually the expected behavior. What do you think? Or am I still missing something?

hofiisek:
Agree. Added in the last commit :)

Randgalt:
Thank you for this

PR #160, opened 2023-12-14
Option to make builder ctor public

### LINKED ISSUE #157: Allow subclassing of generated Builder
I'd like to subclass the generated builder in order to add more quality of life methods, e.g.:

```
public record Grouping(
    UUID id,
    String name,
    String description,
    String context,
    List<String> objectRefs )  {

    public static class Builder extends GroupingBuilder {
        public Builder id(String id) {
            this.id(UUID.fromString(id));
            return this;
        }
    }

}
```

Randgalt:
`@RecordBuilder.Options(publicBuilderConstructors = true)`

Closes #157
Replaces #128


Randgalt:
attn:

- @KangoV
- @toolforger
- @yurybubnov

PR #161, opened 2024-01-01
Support staged builder

### LINKED ISSUE #159: Support staged builder
Something similar to https://immutables.github.io/immutable.html#staged-builder:

> A staged builder is a compile-time safe way to ensure that all required attributes are set. The API, composed of stage interfaces, forces initialization of mandatory attributes in stages, one by one, guiding via code-completion and making it impossible to even call build() before all are set.

Randgalt:
Closes #159

New option to have standard builder, staged builder or both


reda-alaoui (approved):
Amazing ! Thank you 🙏

PR #163, opened 2024-01-02
Support record component initializers

### LINKED ISSUE #110: Defaults
I'd like a way to supply default values, so the record is in the proper state if a field is not set using the builder.

Suggestions:
@Default(myvalueHere)

A magic static method that gets called prior to building with the builder as a parameter? This way we could do validation prior to building as well.

Randgalt:
New component annotation that specifies the name of either a public static final field or a public static method that will be used to initialize each record component in the generated builder so as to support default values.

Closes #110


Randgalt:
@mads-b @pragmasoft-ua @agentgt any comments before I merge?

agentgt:
LGTM. You addressed my concerns of external defaults nicely.

PR #169, opened 2024-02-27
Implement parameterized builder

epkugelmass:
https://github.com/Randgalt/record-builder/issues/157#issuecomment-1963246378


record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -66,9 +69,28 @@ class InternalRecordBuilderProcessor {
...
         packageName = packageNameOpt.orElse(recordActualPackage);
-        builderClassType = ElementUtils.getClassType(packageName,
-                getBuilderName(record, metaData, recordClassType, metaData.suffix()), record.getTypeParameters());
-        typeVariables = record.getTypeParameters().stream().map(TypeVariableName::get).collect(Collectors.toList());
+
+        String builderName = getBuilderName(record, metaData, recordClassType, metaData.suffix());
+        recordTypeVariables = toTypeVariableNames(record.getTypeParameters());
+        List<TypeVariableName> builderTypeParams = new ArrayList<>(recordTypeVariables);
+        List<TypeVariableName> builderTypeParamsWildcard = new ArrayList<>(recordTypeVariables);
+        var self = TypeVariableName.get("SELF");
```
Randgalt, line 77: `SELF` should be defined in `RecordBuilder.Options` metadata in case some client code already uses this name.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -66,9 +69,28 @@ class InternalRecordBuilderProcessor {
         this.metaData = metaData;
         recordClassType = ElementUtils.getClassType(record, record.getTypeParameters());
         packageName = packageNameOpt.orElse(recordActualPackage);
-        builderClassType = ElementUtils.getClassType(packageName,
-                getBuilderName(record, metaData, recordClassType, metaData.suffix()), record.getTypeParameters());
-        typeVariables = record.getTypeParameters().stream().map(TypeVariableName::get).collect(Collectors.toList());
+
+        String builderName = getBuilderName(record, metaData, recordClassType, metaData.suffix());
+        recordTypeVariables = toTypeVariableNames(record.getTypeParameters());
+        List<TypeVariableName> builderTypeParams = new ArrayList<>(recordTypeVariables);
```
Randgalt, line 75: The construction of `builderTypeParams` and `builderTypeParamsWildcard` shouldn't result in a modifiable `ArrayList`. Please move to separate method or `ElementUtils` so that the constructed type is immutable.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
Randgalt, line 1: A general comment. Every generated sub-class should have the same Parameterized Type list as the builder but this PR doesn't do that. It looks like `builderTypeParamsWildcard` does that instead? Why not do what was done before: i.e. all the nested Stage/With classes should have the exact same Parameterized Type list as the builder.

record-builder-test/src/test/java/io/soabase/recordbuilder/test/TestParameterizedBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+
+class TestParameterizedBuilder {
+    @Test
+    void testNoSubclass() {
+        ParameterizedBuilderBuilder<?> builder = ParameterizedBuilderBuilder.builder();
```
Randgalt, line 29: It feels odd to have a static builder in a parameterized builder. It should likely not be added for parameterized and force use of the constructors.

Randgalt: Is the no-subclass even useful? We could make the builder `abstract` when parameterized.

PR #175, opened 2024-03-27
Deconstructors/Tuple generator

### LINKED ISSUE #139: Cannot invoke "javax.lang.model.element.ExecutableElement.getAnnotationMirrors()" because the return value of "javax.lang.model.element.RecordComponentElement.getAccessor()" is null

Randgalt:
This is a speculative PR. Open for discussion.

This is a variation of `@RecordInterface` that generates a general purpose tuple from any class. The impetus for this is to take advantage of Java 21's record patterns with non-record objects. In the future, Java will generally support deconstructors, but it doesn't today. Only real `record`s are supported for record patterns. With the addition here of `@RecordDeconstructor` and object can be use for record pattern matching.

Example:

Given a legacy or non-record class like this:

```
public class BasicDeconstruct {
    private final String name;
    private final int qty;

    public BasicDeconstruct(String name, int qty) {
        this.name = name;
        this.qty = qty;
    }

    public String getName() {
        return name;
    }

    public int getQty() {
        return qty;
    }
}
```

We can now create a tuple-like record from this class by adding annotations:

```
@RecordDeconstructor
public class BasicDeconstruct {
    private final String name;
    private final int qty;

    public BasicDeconstruct(String name, int qty) {
        this.name = name;
        this.qty = qty;
    }

    @RecordDeconstructor.Component
    public String getName() {
        return name;
    }

    @RecordDeconstructor.Component
    public int getQty() {
        return qty;
    }
}
```

This generates the following record:

```
public record BasicDeconstructTuple(String name, int qty) {
    public static BasicDeconstructTuple from(BasicDeconstruct from) {
        return new BasicDeconstructTuple(from.getName(), from.getQty());
    }
}
```

Which can be used for pattern matching:

```
public void foo(BasicDeconstruct instance) {
    switch (BasicDeconstructTuple.from(instance)) {
        case BasicDeconstructTuple(var name, var qty) when name.equals("something") -> ....
    }
}
```

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -74,6 +74,12 @@ public @interface RecordBuilder {
         String interfaceSuffix() default "Record";
 
         /**
+         * Used by {@code RecordTuple}. The generated record will have the same name as the annotated element plus this
+         * suffix. E.g. if the element name is "Foo", the record will be named "FooTuple".
+         */
+        String tupleSuffix() default "Tuple";
+
+        /**
          * The name to use for the copy builder
          */
         String copyMethodName() default "builder";
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordTuple.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordTuple.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 The original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.soabase.recordbuilder.core;
+
+import java.lang.annotation.*;
+
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface RecordTuple {
+    @Retention(RetentionPolicy.SOURCE)
+    @Target(ElementType.METHOD)
+    @interface Component {
+        String value() default "";
+    }
+}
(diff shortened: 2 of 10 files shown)


koppor:
Can I also put the annotation "externally" to an existing class? I think, this is useful when using an external library, where the source should not be modified. -- If I can modify the source, I could use IntelliJ's refactoring "convert to record". OK, that would introduce API incompatibility. Then, the presented approach is good. However, I currently think, for self-maintained code this approach feels to require too much maintenance effort in comparison to the gain. 😅

PR #179, opened 2024-04-10
Fix staged builders for single component records

### LINKED ISSUE #177: Staged builders are not generated for records having only a single field
This is problematic because the main reason you want to use staged builders is so you get a compile-time error if you add a (non-nullable) field to the record, but forget to specify it in the builder. This can happen as easily when you add a second field as it can when you add a third, fourth, or first one for that matter. Staged builders should be generated in all cases.

Randgalt:
Fixes #177

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -196,10 +196,6 @@ class InternalRecordBuilderProcessor {
     }
 
     private void addStagedBuilderClasses() {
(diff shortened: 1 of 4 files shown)


Randgalt:
attn: @gabrielshanahan

gabrielshanahan:
Thanks @Randgalt, LGTM! But I would still advocate for generating staged builders even for 0-component record classes. I know that's a slightly weird and probably rare edge-case, but the same logic as described in #177 still applies - you can add a field at any time, and then want the compiler to tell you what you need to change.

Randgalt:
@gabrielshanahan - I changed to allow no components as well. Please check.

gabrielshanahan:
Awesome, thanks!

PR #180, opened 2024-04-11
Alternate staged builders that handle optional, etc.

### LINKED ISSUE #170: Allow nullable arguments to be optional in StagedBuilders, allowing them to initialize during the final stage
First of all, thank you for the [Staged builders](https://github.com/Randgalt/record-builder/pull/161) support! It helps us a lot to write null-safe code :)

For full adoption in our codebase though, we miss one specific convenience feature that doesn't add any runtime value, but helps tremendously in DX:

The original ["reference" implementation](https://immutables.github.io/immutable.html#staged-builder) supports the following:

> All remaining optional, nullable, and collection attributes can be initialized on a final stage in any order.

I believe it would be a perfect feature in combination with the already existing `interpretNotNulls` option, combining both the safety and ease of use of the builder pattern.

Randgalt:
A variant of staged builders is available that only stages required record components. Any optional components (when `addConcreteSettersForOptional` is enabled) are not staged and are added to the final stage. Additionally, if Collection options are enabled, those too are added to the final stage.

Closes #170

attn: @olsavmic 

Reviewers - please test various use cases. I'm not very familiar with this feature and did what I thought was right.

### DIFF
--- a/options.md
+++ b/options.md
@@ -74,6 +74,12 @@ Use `@RecordBuilder.Options(builderMode = BuilderMode.STAGED)` or `@RecordBuilde
 builders. Staged builders require that each record component is built in order and that each component is specified. The generated builder ensures
 this via individual staged builders. See [TestStagedBuilder](record-builder-test/src/test/java/io/soabase/recordbuilder/test/staged/TestStagedBuilder.java) for examples.
 
+A variant of staged builders is available that only stages required record components. Use `BuilderMode.STAGED_REQUIRED_ONLY` or `BuilderMode.STANDARD_AND_STAGED_REQUIRED_ONLY`. 
+The following are not staged and are added to the final stage:
+- optional components (when `addConcreteSettersForOptional` is enabled)  
+- Any collections matching enabled [Collection options](#collections)  
+- Any annotated compontents that match the `nullablePattern()` pattern option (e.g. `@Nullable`)
+
 ## Default Values / Initializers
 
 | option                                                             | details                                                                                                                          |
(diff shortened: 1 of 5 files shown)


olsavmic:
This is perfect! I'm going to test it on our codebase and report possible issues in the next few days.

I don't expect any though, this is the design I envisioned, thank you!

Randgalt:
Let me know and I'll merge.

olsavmic:
There is a bug when immutableCollections are enabled together with STAGED_REQUIRED_ONLY option. The code does not compile, as the stage for the collection is detected as NOT REQUIRED yet the builder is not generated.

See https://github.com/Randgalt/record-builder/pull/182 for a reproducer.

olsavmic:
Fix for the bug is in https://github.com/Randgalt/record-builder/pull/182/commits/9632f4133b36431db5b5fa217426b0448a4fca60.

---

The configurations around collections and nullability are getting pretty complicated :/
I'll test the different combinations, but I'll mainly try to come up with a description of the sane configurations.

`useImmutableCollections`, `useUnmodifiableCollections` (basically equivalent)
`allowNullableCollections`
`interpretNotNulls`
`nullablePattern`

olsavmic:
The changes in https://github.com/Randgalt/record-builder/pull/182 make it work for us, but as I said above, the behaviour with nullability is now not uniform.

What I'd like to see is the following:

"1st" layer:

- Define whether fields are `@Null` by default or `@NotNull` by default - that information is then combined together with annotations, allowing them to override the defaults.
    - @Null by default, @Null field -> Null
    - @Null by default, @NotNull field -> NotNull
    - @NotNull by default, @Null field -> Nullable
    - @NotNull by default, @NotNull -> NotNull

"2nd" layer:
- `interpretNotNulls` is configuring just the runtime checks behavior
    - When a field is resolved as NotNull, a runtime check is generated (regardless of whether it's a collection or not)
  
- `allowNullableCollections` 
    - `false` -> Automatically set to empty collection
    - `true` -> may be NULL (but the runtime check still applies, meaning that NotNull field would throw if it's not set and `interpretNotNulls` is true)

- Anytime field is resolved as NotNull, it is **required** in the STAGED_REQUIRED_ONLY

Randgalt:
@olsavmic I agree some of the options have gotten out of hand. Too many cooks and not enough quality. I'd like to deprecate a lot of it and re-work it.

PR #185, opened 2024-07-29
Adds RecordBuilder.Options.defaultNotNull parameters

agustafson:
Thanks for putting together record-builder! It's definitely a welcome addition and enhancement on top of `record`s.

### Issue/Enhancement
Some codebases expect fields to be non-null by default, for example if NullAway enforcement is in place. Therefore it is useful to set a parameter to expect that all fields will be non-null unless explicitly marked as nullable.
Currently our codebase uses https://immutables.github.io/immutable.html which defaults to non-null fields, and it would be good to have record-builder have the capability for the same behaviour.

### Changes
Adds the ability to default fields to be non-null, unless they are annoted with a nullable annotation, ie:
```java
@RecordBuilder.Options(defaultNotNull = true/false)
```

### Example
For example:
```java
@RecordBuilder.Options(defaultNotNull = true)
@RecordBuilder
public record RecordWithDefaultNotNull(Integer notNullInteger, String notNullString, @Null String nullString) {
}
```
will check/ensure that `notNullInteger` and `notNullString` are not null via `requireNonNull` calls.

### Questions
- What happens for collections? Can we enforce non-nullability in the same way by defaulting to non-null? Or is this not required because collections are non-null by default anyway.
- Can we also support other annotations such as `javax.annotation.ParametersAreNonnullByDefault`? That might be a future enhancement.

### DIFF
--- a/options.md
+++ b/options.md
@@ -94,12 +94,13 @@ for an example.
 
 ## Null Handling
 
-| option                                                          | details                                                                                                          |
-|-----------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------|
-| `@RecordBuilder.Options(interpretNotNulls = true/false)`        | Add not-null checks for record components annotated with any null-pattern annotation. The default is `false`.    |
-| `@RecordBuilder.Options(interpretNotNullsPattern = "regex")`    | The regex pattern used to determine if an annotation name means non-null.                                        |
-| `@RecordBuilder.Options(allowNullableCollections = true/false)` | Adds special null handling for record collectioncomponents. The default is `false`.                              |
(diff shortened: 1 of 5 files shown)


agustafson:
Hi @Randgalt just bumping this one to see if there's any progress please? I'm happy to rework it if there's any suggestions that you have. Thanks!

Randgalt:
I apologize - I've been completely slammed at work. I will get to this later today or tomorrow.

agustafson:
Don't apologise! It's tough work juggling work and open-source projects, so I appreciate you taking some time to look at it.

Randgalt:
Good change - thank you

agustafson:
Amazing! Thanks for that.

PR #194, opened 2024-11-13
Add annotationProcessorPaths to the pom.xml code snippet

mk868:
As of JDK23, javac doesn't automatically detect discoverable annotation processors by default.
In this PR, I add the explicit definition of the processor annotation in the maven-compiler-plugin.
I hope this will make it easier for future users to add RecordBuilder to their project.

Sources:

- [Changes Default Annotation Processing Policy](https://inside.java/2024/06/18/quality-heads-up/#jdk-23-changes-default-annotation-processing-policy)

**Edit**
Worth to add: for Gradle, the annotation processor is already explicitly defined via `annotationProcessor`, so there is no problem with JDK23.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,15 +327,35 @@ annotation. Use `packagePattern` to change this (see Javadoc for details).
 
 ### Maven
 
-Add a dependency that contains the discoverable annotation processor:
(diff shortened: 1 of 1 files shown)


Randgalt:
Thank you.

Issue #189, opened 2024-10-23
JSR303 @Valid is not getting filtered out


elyse-yao:
I have the record:
`public record Foo(String id, @Valid @NotNull Bar properties)
        implements FooBuilder.With {}`

And I'm getting this error:
`A method return value must not be marked for cascaded validation more than once in a class hierarchy, but the following two methods are marked as such: With#properties(), Foo#properties()`

I saw in the [record-builder processor source](https://github.com/Randgalt/record-builder/blob/c168f33dda2c5a00e7a0b80d510477af8c1570b7/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java#L866) that it's meant to be filtered out, but that only applies to the `javax` version of the annotation, not the `jakarta.validation.Valid` used in spring. I know I can just disable all annotation inheritance, but Is there a way to configure record-builder to filter out specific annotations instead?

Randgalt:
Right - the `jakarta` rename mess. We should update for that and filter both. I'd appreciate a PR with this change.

elyse-yao:
(^ @Randgalt small bump for the PR, sorry for the delay!)

PR #195, opened 2024-11-14
Filter out jakarta @Valid annotations

elyse-yao:
This modifies the record processor to filter out `@Valid` annotations from both javax/jakarta, which fixes https://github.com/Randgalt/record-builder/issues/189. I also added tests for both.

Note: this doesn't add support for validation using Jakarta, since that would require bumping the Hibernator version and changes to the validator/additional validators to preserve support for both validation APIs.


Randgalt:
Please rebase on master and sqaush

PR #197, opened 2025-01-22
Consider initializers for staged required only builders

### LINKED ISSUE #187: Default values and staged builders
Staged builders are useful to force compile time safety that all required attributes are set. The autocomplete they give is lovely too.

If any components are marked as nullable then they are optional.

The same isn't true for components marked with a default value. You're forced to specify a value. This is wrong I feel.

Sample code to see:

```
@MyStagedRecordBuilder
public record Person(
        @Initializer("DEFAULT_NAME") String name,
        @Nullable int age,
        String somethingElse
) {

    public static final String DEFAULT_NAME = "John Doe";

    void test() {
        var x = PersonBuilder.staged()
                .name("ds")
                .somethingElse("aada")
                .age(21);
    }
}
```

In this instance as name has a default value the only component I should be forced to specify is `somethingElse`, but i'm forced to specify `name`.

I'm conscious at this point this is trying to turn the library into a replacement for named and default arguments found in other languages. But seeing as both of these pieces of functionality are available I thought it worth raising :)

frederikb:
### Background

Fixes #187

### Changes

Components with initializers (annotated with `@RecordBuilder.Initializer(...)`) are now excluded from staging in "staged required only" builders and are added to the final stage.

A new option `skipStagingForInitializedComponents` controls this behavior. 
By default, initialized components are still considered required and staged.

### Discussion Points

- [x] This is a breaking change to "staged required only" builders. Is this acceptable or would this need to be a behavior toggled through a new option? - **Added option to control this behavior**

- [ ] This is my first contribution to this project and the first time taking a look at the codebase. Is this minimal change correctly done or did I miss some edge cases?

### DIFF
--- a/options.md
+++ b/options.md
@@ -78,13 +78,15 @@ A variant of staged builders is available that only stages required record compo
 The following are not staged and are added to the final stage:
 - optional components (when `addConcreteSettersForOptional` is enabled)  
 - Any collections matching enabled [Collection options](#collections)  
-- Any annotated compontents that match the `nullablePattern()` pattern option (e.g. `@Nullable`)
+- Any annotated components that match the `nullablePattern()` pattern option (e.g. `@Nullable`)
+- Any components with initializers (annotated with `@RecordBuilder.Initializer(...)`) when `skipStagingForInitializedComponents` is enabled
 
 ## Default Values / Initializers
 
-| option                                                             | details                                                                                                                          |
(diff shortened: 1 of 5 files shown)


Randgalt:
Do you mind adding an option for this? I really don't want to break any code.

frederikb:
@Randgalt I added an option and disabled the new behavior by default. Could you please take another look?

Randgalt:
Thank you for the PR

PR #201, opened 2025-02-10
[Fixes #200] Fix for the issue with addFunctionalMethodsToWith and getter/boolean prefix settings not playing nice with each other.

### LINKED ISSUE #200: Using the getterPrefix and/or booleanPrefix settings with the addFunctionalMethodsToWith setting generates a broken nested "With" interface.
Using the following versions:
- RecordBuilder: 44
- Java: 17/21

Minimal example to reproduce the issue:
```java
@RecordBuilder
@RecordBuilder.Options(getterPrefix = "get",
                       booleanPrefix = "is",
                       addFunctionalMethodsToWith = true)
public record Minimal(int value, boolean flag) implements MinimalBuilder.With {}
```

The above code produces the following functional interfaces:
```java
/**
 * Map record components into a new object
 */
default <R> R map(Function<R> proc) {
    return proc.apply(getValue(), isFlag());
}

/**
 * Perform an operation on record components
 */
default void accept(Consumer proc) {
    proc.apply(getValue(), isFlag());
}
```
**Edit:** I wasn't clear enough about why the generated code above is a bad thing. The `With` interface that these methods are created in does not have either of the `getValue()` or `isFlag()` methods. Which causes the build to break for the generated class.

Found this at random while poking around in the code experimenting with things. I also have a potential fix for this that I will create a PR for shortly.

matt-isaman:
This is, I believe, the minimal change needed to fix issue#200 (other than the .gitignore changes of course).

That said, while poking around after making my change I noticed that the `boolean` argument to the `addComponentCallsAsArguments` method is now _always_ set to `false`. Assuming there are no objections to this change I could also update this PR to remove that parameter if desired. From what my IDE tells me there are only three places where this method is used so this additional change would also be quite small.

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -378,7 +378,7 @@ class InternalRecordBuilderProcessor {
          */
         var codeBlockBuilder = CodeBlock.builder().add("return new $L$L(", builderClassType.name(),
                 typeVariables.isEmpty() ? "" : "<>");
-        addComponentCallsAsArguments(-1, codeBlockBuilder, false);
+        addComponentCallsAsArguments(-1, codeBlockBuilder);
         codeBlockBuilder.add(");");
         var methodSpec = MethodSpec.methodBuilder(metaData.withClassMethodPrefix())
                 .addAnnotation(generatedRecordBuilderAnnotation)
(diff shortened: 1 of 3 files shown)


Randgalt:
It appears that `addComponentCallsAsArguments` is only used by the `With` now and the prefix handling should never be applied.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -1118,7 +1118,7 @@ private MethodSpec buildFunctionalHandler(String className, String methodName, b
             methodBuilder.addJavadoc("Perform an operation on record components");
         }
         codeBlockBuilder.add("proc.apply(");
-        addComponentCallsAsArguments(-1, codeBlockBuilder, true);
+        addComponentCallsAsArguments(-1, codeBlockBuilder, false);
```
Randgalt, line 1121: The `usePrefixedName` argument should be removed. It will always be `false`.

matt-isaman: I have updated the PR to remove the parameter.

Randgalt:
Thank you!

PR #206, opened 2025-04-09
Add support for a builder with required components

fredyw:
When the record components are annotated with `@RecordBuilder.Required`, the builder method will take the required record components as its parameters. This is particularly useful to ensure that the builder is always populated with the required fields at compile-time instead of relying on a runtime validation.

Example:
```
public record MyRecord(@Required int a, @Required int b, int c)

public class RequiredComponentsBuilder {
    ...
    public static MyRecordBuilder builder(int a, int b) {
        return new MyRecordBuilder().a(a).b(b);
    }
    ...
}
```


Randgalt:
If the arguments/parameters are required, we should not include the no-arg `builder()`. Or, there should be some kind of validation elsewhere. 

tbh - I'm not sure I'm in favor of this. Anyone else?

fredyw:
> If the arguments/parameters are required, we should not include the no-arg `builder()`. Or, there should be some kind of validation elsewhere.
> 
> tbh - I'm not sure I'm in favor of this. Anyone else?

I agree that the no-arg `builder()` should not be included or else it defeats the purpose of this change. I updated the PR to remove it when `@Required` is present. Let me know what you think.

PR #221, opened 2025-05-01
Fixed #220 - Removed unused maven-shade-plugin

### LINKED ISSUE #220: Why is maven-shade-plugin declared?
In the `pom.xml` there is the maven-shade-plugin defined, but it is not used anywhere?

khmarbaise:


Randgalt:
Thank you

PR #223, opened 2025-05-01
Fixed #222 - Separation pluginManagement/plugins

### LINKED ISSUE #222: Separation pluginManagement/plugins
* No life cycle binding in pluginManagement
* Remove the plugins which are bound by the [life cycles](https://maven.apache.org/ref/3.9.9/maven-core/default-bindings.html)

khmarbaise:
- Moved life cycle binding to plugins instead of pluginManagement.
 - Removed other plugins which are bound by default via life cycle.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,14 +203,6 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-source-plugin</artifactId>
                     <version>${maven-source-plugin-version}</version>
-                    <executions>
-                        <execution>
-                            <id>attach-sources</id>
-                            <goals>
-                                <goal>jar-no-fork</goal>
-                            </goals>
-                        </execution>
-                    </executions>
                 </plugin>
 
                 <plugin>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -328,61 +311,63 @@
...
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-install-plugin</artifactId>
-            </plugin>
-
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-source-plugin</artifactId>
+                <executions>
```
Randgalt, line 332: I've always put this in the pluginManagement. Why should it be here out of curiosity?

khmarbaise: Technically it is working in pluginManagement, but I recommend to separate the configuration/version definition of plugins... in dependencyManagement... and in build do the bindings to a life cycle....

Randgalt: For me, having it in pluginManagement is more, well, manageable :D I prefer to leave it there.

PR #224, opened 2025-05-05
Specifying nullability annotation for concrete setters of optional

### LINKED ISSUE #203: Missing nullability annotation when using the addConcreteSettersForOptional setting
When using record-builder with nullaway, the record-builder approach of keeping field annotations works nicely for nullable fields. In our code base, we often use fields wrapped with `Optional` instead of nullable fields though, which of course don't have a `@Nullable` annotation. This causes issues when enabling the option `addConcreteSettersForOptional` for nullable convenience setters, as those don't have a `@Nullable` annotation either (which nullaway then complains about). It would be great if we could provide the nullability annotation to use for such cases via the config and then have it on the generated setters.

zoehneto:
Fixes #203 by adding a new optional config option for the desired nullability annotation.

### DIFF
--- a/options.md
+++ b/options.md
@@ -56,19 +56,20 @@ The names used for generated methods, classes, etc. can be changed via the follo
 
 ## Miscellaneous
 
-| option                                                               | details                                                                                                                                                          |
-|----------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `@RecordBuilder.Options(inheritComponentAnnotations = true/false)`   | If true, any annotations (if applicable) on record components are copied to the builder methods. The default is `true`.                                          |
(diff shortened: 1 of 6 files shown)


dosomder:
Nice job, this would be very useful. @Randgalt any news on this?

Randgalt:
I'll try to get to this soon

Randgalt:
@zoehneto I apologize it took me so long on this - I've been slammed at work. I'm just now picking my head up...

I've just pushed a change that I prefer. LMKWYK:

- Generalize the nullable annotation option so it can be used for other things
- Have `addConcreteSettersForOptional` be an enum that has three choices: off, on and on with nullable

zoehneto:
@Randgalt thanks for looking at the PR! I really like the changes. Good to merge for me.

PR #226, opened 2025-05-14
Stream and From should not be considered for dup name check

### LINKED ISSUE #225: #213 breaks some code
> @Randgalt, I'm checking out the 46th release (updated from v44). Some of my records broke because of the issue. The minimal record looks like this:
> ```java
> @RecordBuilder.Template(
>     options =
>         @RecordBuilder.Options(
>             enableGetters = false,
>             enableWither = false,
>             fileIndent = "  ",
>             inheritComponentAnnotations = false,
>             prefixEnclosingClassNames = false))
> @Retention(RetentionPolicy.SOURCE)
> @Target(ElementType.TYPE)
> @Inherited
> public @interface Builder {}
> 
> @Builder
> record ConvertRequest(Amount from, Currency to ....) {}
> ```
> So the `from` component conflicts with `fromMethodName`. I see what's broken, but I can't understand why that is so.
> * My record doesn't use wither, so no `from`-methods are generated at all.
> * If I provide `fromMethodName = "__from__"` and enable withers, I don't see any possible conflicts too. Yes, there are two methods named `from`, but one of them is a static method, another is an instance one.
> 
> I can stick with my workaround (using `__from__` for generated method name), but I'm curious about the motivation of adding these checks. 

 _Originally posted by @genuss in [#213](https://github.com/Randgalt/record-builder/issues/213#issuecomment-2880230897)_

Randgalt:
Fixes #225

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -174,16 +174,6 @@ class InternalRecordBuilderProcessor {
                 return false;
             }
 
-            if (recordComponent.name().equals(metaData.fromMethodName())) {
-                reportError.accept(recordComponent.name(), "fromMethodName");
-                return false;
-            }
-
-            if (recordComponent.name().equals(metaData.componentsMethodName())) {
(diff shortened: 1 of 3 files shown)


genuss:
Just checked on my project: everything works fine! You're the fastest 😄

Randgalt:
Not always ;D I just happened to have some time.

PR #227, opened 2025-05-14
In some JDKs the accessors list can be null

Randgalt:
Possible fix for #139


gardellajuanpablomodusbox:
Let me check this branch.

PR #232, opened 2025-06-03
Add support to configure adding jackson @JsonPOJOBuilder annotations to generated builders

### LINKED ISSUE #229: Make RecordBuilder Jackson-friendly
I'd like to use RecordBuilder in combination with Jackson using the builder for deserialization in order to leverage immutable collections and default values/initializers. This should work with a vanilla `ObjectMapper` instance without any custom mixins/configuration.

### Problem description

Starting from the following test, it fails as `type` and `properties` is `null` as it does not use the builder at all (expected):

```java
package com.example.model;

import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;
import io.soabase.recordbuilder.core.RecordBuilder;
import org.junit.jupiter.api.Test;

import java.util.Map;

import static org.assertj.core.api.Assertions.assertThat;

class RecordSerializationTest {

    @RecordBuilder
    @RecordBuilder.Options(
        useImmutableCollections = true
    )
    // @JsonDeserialize(builder = RecordSerializationTestMyTestModelBuilder.class)
    public record MyTestModel(
        String name,
        @RecordBuilder.Initializer("DEFAULT_TYPE") String type,
        Map<String, Object> properties
    ) {
        public static final String DEFAULT_TYPE = "dummy";
    }

    private final ObjectMapper objectMapper = new ObjectMapper();

    @Test
    void deserializingModelInvokesBuilder() throws JsonProcessingException {
        final var json = """
            {
              "name" : "test"
            }
            """;

        final var model = objectMapper.readValue(json, MyTestModel.class);
        assertThat(model.name()).isEqualTo("test");
        assertThat(model.type()).isEqualTo("dummy");
        assertThat(model.properties()).isNotNull().isEmpty();
    }
}
```

When uncommenting the `@JsonDeserialize` line, it tries to use the builder, but fails as Jackson's default implementation expects builder methods to start with `with`:

```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "name" (class com.example.model.RecordSerializationTestMyTestModelBuilder), not marked as ignorable (0 known properties: ])
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 2, column: 13] (through reference chain: com.example.model.RecordSerializationTestMyTestModelBuilder["name"])
```

A solution to reconfigure Jackson to be aware of the builder using no prefix would be to annotate the builder class with `@JsonPOJOBuilder(withPrefix = "")`, but either I missed it or it is currently not possible to apply this to the generated builder class.

An ugly workaround is to enable public constructors on the builder class + to extend the generated builder with an annotated one (this makes the test pass):

```java
    @RecordBuilder
    @RecordBuilder.Options(
        useImmutableCollections = true,
        publicBuilderConstructors = true
    )
    @JsonDeserialize(builder = MyTestModel.AnnotatedBuilder.class)
    public record MyTestModel(
        String name,
        @RecordBuilder.Initializer("DEFAULT_TYPE") String type,
        Map<String, Object> properties
    ) {
        public static final String DEFAULT_TYPE = "dummy";

        @JsonPOJOBuilder(withPrefix = "")
        public static class AnnotatedBuilder extends RecordSerializationTestMyTestModelBuilder {
        }
    }
```

### Possible solutions

It would be great if RecordBuilder could expose a way to make this work nice together with Jackson without any workarounds or reconfiguration of the ObjectMapper. I'm aware that this affects interoperability with a third-party library but as this is a very common use-case (e.g. deserialization in Spring Boot controllers) it would be great to have built-in Jackson support directly in RecordBuilder.

I could imagine multiple ways of doing this:

- Add an option like `addJsonPOJOBuilderAnnotation` and add `@JsonPOJOBuilder(withPrefix = "")` to the generated builder if configured (potentially auto-enable this when Jackson is found on the classpath?).
- Kind of a workaround: ~provide a way to configure the setter prefix for builder methods to it can be set to `with`.~ I noticed this is already there, but not documented - created a small PR to update the docs in https://github.com/Randgalt/record-builder/pull/228. This works when setting it to `with` but it would be nicer to keep the builders as they are instead of introducing the prefix.
- Provide a way to specify arbitrary annotations which should be applied to the builder (given technical feasibility, as far as I've seen there have been similar discussions before).

maff:
Adds a new option structure to instruct record-builder to add a Jackson `@JsonPOJOBuilder` annotation to the generated builder. While this implements library-specifics, it does not add Jackson as dependency (only on the test module).

**Example usage:**

Auto-detect Jackson version (default)
```java
@RecordBuilder.Options(jackson = @RecordBuilder.JacksonConfig(jsonPOJOBuilder = true))
```

Explicitly add Jackson 2.x annotation (fail if J2 is not available):
```java
@RecordBuilder.Options(
    jackson = @RecordBuilder.JacksonConfig(
        jsonPOJOBuilder = true,
        version = JacksonVersion.JACKSON_2
    )
)
```

With custom setter prefix:
```java
@RecordBuilder.Options(
    jackson = @RecordBuilder.JacksonConfig(jsonPOJOBuilder = true),
    setterPrefix = "set"
)
```

Closes #229

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,8 @@
         <hibernate-validator-version>6.2.0.Final</hibernate-validator-version>
         <jakarta-validation-api-version>3.1.0</jakarta-validation-api-version>
         <javax-el-version>3.0.1-b09</javax-el-version>
-        <jackson-version>2.19.0</jackson-version>
+        <jackson2-version>2.21.0</jackson2-version>
+        <jackson3-version>3.0.4</jackson3-version>
         <central-publishing-maven-plugin-version>0.7.0</central-publishing-maven-plugin-version>
         <jspecify-version>1.0.0</jspecify-version>
         <lombok-version>1.18.42</lombok-version>
@@ -171,7 +172,12 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
-                <version>${jackson-version}</version>
+                <version>${jackson2-version}</version>
+            </dependency>
+            <dependency>
+                <groupId>tools.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>${jackson3-version}</version>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
         <hibernate-validator-version>6.2.0.Final</hibernate-validator-version>
         <jakarta-validation-api-version>3.1.0</jakarta-validation-api-version>
         <javax-el-version>3.0.1-b09</javax-el-version>
+        <jackson-version>2.19.0</jackson-version>
         <central-publishing-maven-plugin-version>0.7.0</central-publishing-maven-plugin-version>
         <jspecify-version>1.0.0</jspecify-version>
         <lombok-version>1.18.42</lombok-version>
@@ -165,6 +166,12 @@
                 <groupId>org.hibernate.validator</groupId>
                 <artifactId>hibernate-validator</artifactId>
                 <version>${hibernate-validator-version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>${jackson-version}</version>
             </dependency>
 
             <dependency>
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -359,6 +359,8 @@ public @interface RecordBuilder {
          * @see #nullablePattern
          */
         boolean defaultNotNull() default false;
+
+        boolean addJacksonAnnotations() default false;
     }
 
     @Retention(RetentionPolicy.CLASS)
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -360,7 +360,30 @@ public @interface RecordBuilder {
          */
         boolean defaultNotNull() default false;
 
-        boolean addJacksonAnnotations() default false;
+        /**
+         * Configuration for Jackson annotation support on generated builders.
+         *
+         * @see JacksonConfig
+         */
+        JacksonConfig jackson() default @JacksonConfig;
+    }
+
+    /**
+     * Configuration for Jackson annotation support on generated builders.
+     */
+    @Retention(RetentionPolicy.CLASS)
+    @Target(ElementType.ANNOTATION_TYPE)
+    @interface JacksonConfig {
+        /**
+         * Add {@code @JsonPOJOBuilder} annotation to generated builder. This annotation works with
+         * {@code @JsonDeserialize(builder = ...)} on the record.
+         */
+        boolean jsonPOJOBuilder() default false;
+
+        /**
+         * Which Jackson version to use for annotations.
+         */
+        JacksonVersion version() default JacksonVersion.AUTO;
     }
 
     @Retention(RetentionPolicy.CLASS)
@@ -378,6 +401,29 @@ public @interface RecordBuilder {
 
     enum ConcreteSettersForOptionalMode {
         DISABLED, ENABLED, ENABLED_WITH_NULLABLE_ANNOTATION,
+    }
+
+    /**
+     * Specifies which Jackson version(s) to use when generating builder annotations.
+     */
+    enum JacksonVersion {
+        /**
+         * Automatically detect Jackson version(s) on classpath and add all found annotations. If both Jackson 2.x and
+         * 3.x are present, both annotations will be added.
+         */
+        AUTO,
+
+        /**
+         * Only add Jackson 2.x annotations (com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder). Fails if
+         * Jackson 2.x is not found on classpath.
+         */
+        JACKSON_2,
+
+        /**
+         * Only add Jackson 3.x annotations (tools.jackson.databind.annotation.JsonPOJOBuilder). Fails if Jackson 3.x is
+         * not found on classpath.
+         */
+        JACKSON_3,
     }
 
     /**
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -89,7 +89,7 @@ class InternalRecordBuilderProcessor {
             builder.addAnnotation(recordBuilderGeneratedAnnotation);
         }
 
-        addJacksonAnnotations();
+        new JacksonSupport(processingEnv).addJacksonAnnotations(metaData, builder);
 
         if (!validateMethodNameConflicts(processingEnv, recordFacade.element())) {
             builderType = Optional.empty();
@@ -199,18 +199,6 @@ class InternalRecordBuilderProcessor {
         } else {
             builder.addModifiers(Modifier.PUBLIC);
         }
-    }
-
-    private void addJacksonAnnotations() {
-        if (!metaData.addJacksonAnnotations()) {
-            return;
-        }
-
-        final var annotationSpec = AnnotationSpec
-                .builder(ClassName.get("com.fasterxml.jackson.databind.annotation", "JsonPOJOBuilder"))
-                .addMember("withPrefix", "$S", metaData.setterPrefix()).build();
-
-        builder.addAnnotation(annotationSpec);
     }
 
     private void addOnceOnlySupport() {
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -89,6 +89,8 @@ class InternalRecordBuilderProcessor {
             builder.addAnnotation(recordBuilderGeneratedAnnotation);
         }
 
+        addJacksonAnnotations();
+
         if (!validateMethodNameConflicts(processingEnv, recordFacade.element())) {
             builderType = Optional.empty();
             return;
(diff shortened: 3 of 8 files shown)


maff:
@Randgalt any feedback on this approach? I would be great to have this as an option for projects relying on Jackson and still wanting to benefit from RecordBuilder's initialization & defaults logic.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -196,6 +198,18 @@ private void addVisibility(boolean builderIsInRecordPackage, Set<Modifier> modif
         }
     }
 
+    private void addJacksonAnnotations() {
+        if (!metaData.addJacksonAnnotations()) {
+            return;
+        }
+
+        final var annotationSpec = AnnotationSpec
```
semyon-levin-workato: I'm afraid it's not that simple. Recently, Jackson 3 was released with changed package names. Is it possible to check what annotations are presented on the classpath and add appropriate annotation here?

maff:
@Randgalt @semyon-levin-workato Updated to support both Jackson 2 and 3 by detecting available classes on the classpath, with support to specify a specific version in case both versions are present.

Would be awesome if we could progress with this feature for better out-of-the-box Jackson compatibility.

semyon-levin-workato:
@Randgalt could you please look into this?

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+        case JACKSON_3 -> {
+            if (!jackson3Present) {
+                processingEnv.getMessager().printMessage(ERROR,
+                        "jackson.version is set to JACKSON_3 but Jackson 3.x is not found on classpath. "
+                                + "Add jackson-databind 3.x dependency or change version to AUTO.");
+                return;
+            }
+
+            addJacksonAnnotations(metaData, builder, JACKSON_3_ANNOTATION_PACKAGE);
+        }
```
Randgalt, line 91: Let's remove those returns

```suggestion
        case AUTO -> {
            if (!jackson2Present && !jackson3Present) {
                processingEnv.getMessager().printMessage(ERROR,
                        "jackson.jsonPOJOBuilder is enabled but Jackson is not found on classpath. "
                                + "Add jackson-databind dependency or disable jsonPOJOBuilder.");
            }
            else if (jackson2Present) {
                addJacksonAnnotations(metaData, builder, JACKSON_2_ANNOTATION_PACKAGE);
            }
            else /*if (jackson3Present)*/ {
                addJacksonAnnotations(metaData, builder, JACKSON_3_ANNOTATION_PACKAGE);
            }
        }

        case JACKSON_2 -> {
            if (!jackson2Present) {
                processingEnv.getMessager().printMessage(ERROR,
                        "jackson.version is set to JACKSON_2 but Jackson 2.x is not found on classpath. "
                                + "Add jackson-databind 2.x dependency or change version to AUTO.");
            }
            else {
                addJacksonAnnotations(metaData, builder, JACKSON_2_ANNOTATION_PACKAGE);
            }
        }

        case JACKSON_3 -> {
            if (!jackson3Present) {
                processingEnv.getMessager().printMessage(ERROR,
                        "jackson.version is set to JACKSON_3 but Jackson 3.x is not found on classpath. "
                                + "Add jackson-databind 3.x dependency or change version to AUTO.");
            } else {
                addJacksonAnnotations(metaData, builder, JACKSON_3_ANNOTATION_PACKAGE);
            }
        }
```

pom.xml
```diff
@@ -171,7 +172,12 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
-                <version>${jackson-version}</version>
+                <version>${jackson2-version}</version>
+            </dependency>
+            <dependency>
```
Randgalt, line 177: nit: add a new line here

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
Randgalt, line 1: Please squash the two commits

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -378,6 +403,29 @@ enum ConcreteSettersForOptionalMode {
         DISABLED, ENABLED, ENABLED_WITH_NULLABLE_ANNOTATION,
     }
 
+    /**
+     * Specifies which Jackson version(s) to use when generating builder annotations.
+     */
+    enum JacksonVersion {
+        /**
```
Randgalt, line 410: Let's add a `NONE` attribute here and remove `@JacksonConfig`. I'd also rename this `RecordBuilderJacksonVersion` or something to avoid name collisions with other projects (I realize some other names in here suffer from this).

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -359,6 +359,31 @@
...
+         */
+        JacksonConfig jackson() default @JacksonConfig;
+    }
+
+    /**
+     * Configuration for Jackson annotation support on generated builders.
+     */
+    @Retention(RetentionPolicy.CLASS)
+    @Target(ElementType.ANNOTATION_TYPE)
+    @interface JacksonConfig {
```
Randgalt, line 376: See comment below. We can remove this annotation.

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -359,6 +359,31 @@
          * @see #nullablePattern
          */
         boolean defaultNotNull() default false;
+
+        /**
+         * Configuration for Jackson annotation support on generated builders.
+         *
+         * @see JacksonConfig
+         */
+        JacksonConfig jackson() default @JacksonConfig;
```
Randgalt, line 368:
```suggestion
JacksonVersion jackson() default JacksonVersion.NONE;
```

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+    private final boolean jackson2Present;
+    private final boolean jackson3Present;
+
+    JacksonSupport(ProcessingEnvironment processingEnv) {
+        this.processingEnv = processingEnv;
+        jackson2Present = isAnnotationClassPresent(JACKSON_2_ANNOTATION_PACKAGE, JSON_POJO_BUILDER);
+        jackson3Present = isAnnotationClassPresent(JACKSON_3_ANNOTATION_PACKAGE, JSON_POJO_BUILDER);
+    }
+
+    private boolean isAnnotationClassPresent(String packageName, String className) {
```
Randgalt, line 43: `className` is always `JSON_POJO_BUILDER`

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final String JACKSON_2_ANNOTATION_PACKAGE = "com.fasterxml.jackson.databind.annotation";
+    private static final String JACKSON_3_ANNOTATION_PACKAGE = "tools.jackson.databind.annotation";
+
+    private static final String JSON_POJO_BUILDER = "JsonPOJOBuilder";
+
+    private final ProcessingEnvironment processingEnv;
+    private final boolean jackson2Present;
+    private final boolean jackson3Present;
+
+    JacksonSupport(ProcessingEnvironment processingEnv) {
```
Randgalt, line 37: Given how this is used it would be better as a utility class with a private ctor and the methods becoming `static`.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+        this.processingEnv = processingEnv;
+        jackson2Present = isAnnotationClassPresent(JACKSON_2_ANNOTATION_PACKAGE, JSON_POJO_BUILDER);
+        jackson3Present = isAnnotationClassPresent(JACKSON_3_ANNOTATION_PACKAGE, JSON_POJO_BUILDER);
+    }
+
+    private boolean isAnnotationClassPresent(String packageName, String className) {
+        return processingEnv.getElementUtils().getTypeElement(packageName + "." + className) != null;
+    }
+
+    public void addJacksonAnnotations(RecordBuilder.Options metaData, TypeSpec.Builder builder) {
```
Randgalt, line 47: can be package-private

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+        // return without further processing if no annotation is enabled
+        if (!anyJacksonAnnotationEnabled(metaData)) {
+            return;
+        }
```
Randgalt, line 51: This is unnecessary right?

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/JacksonSupport.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private void addJacksonAnnotations(RecordBuilder.Options metaData, TypeSpec.Builder builder, String packageName) {
+        if (metaData.jackson().jsonPOJOBuilder()) {
+            addJsonPOJOBuilderAnnotation(metaData, builder, packageName);
+        }
+    }
+
+    private void addJsonPOJOBuilderAnnotation(RecordBuilder.Options metaData, TypeSpec.Builder builder,
+            String packageName) {
+        final var annotationSpec = AnnotationSpec.builder(ClassName.get(packageName, JSON_POJO_BUILDER))
```
Randgalt, line 107: We also need to consider `metaData.builderMethodName()` and `metaData.withClassMethodPrefix()` right? `JsonPOJOBuilder` has attributes for that.

Randgalt:
My apologies for how long this has taken but I just haven't had a lot of time recently. I've done a first pass review. In general I'm very positive on this. Thank you.

PR #237, opened 2025-08-10
@Initializer wasn't handling parameterized classes properly

### LINKED ISSUE #235: Defaults for empty list not works as expected
I use the following construction
```java 
@RecordBuilder
public record OutRestV1x0(
    @Initializer(value = "of", source = List.class) List<String> ids
) {}
```
expected builder output is
```java
public class OutRestV1x0Builder {
    private List<String> ids = List.of();
//...
}
```
But I see the info in the log `No matching public static field or method found for initializer named: of` and generated code is
```java
public class OutRestV1x0Builder {
    private List<String> ids;
//...
}
```

Randgalt:
Need to erase types when finding the initializer method and when creating the call to it.

Fixes #235


Randgalt:
attn: https://github.com/foal

PR #248, opened 2025-10-09
Accessor annotations must be supplemented by type annotations

### LINKED ISSUE #111: Support all varieties of Null handling
Yes that will probably be good for all. However I can't keep working on #106 or the other nullable annotation bugs till some decisions are made. Which is why I have been so back and forth on this as there are some serious previous design decisions that are impacting me.

So I'm fine with holding off indefinitely till some decisions are made (like lets copy what immutables does).

And I'm sorry for coming off rough and flooding the inbox but I have already sunken some time on this (a lot more than just changing some call from `of` to `ofNullable`) as I sort of felt obligated given previous discussions on reddit and whatnot.

I'll stop work and commenting till the dust clears.

_Originally posted by @agentgt in https://github.com/Randgalt/record-builder/issues/107#issuecomment-1104038505_

Randgalt:
It appears that calling `component.getAccessor().getAnnotationMirrors()` on a record component accessor does not return all annotations for some reason. This was determined by debugging and trial-and-error.

The annotations returned for the accessor must be supplemented with annotations from the `asType()` called on the component. Note: the annotations from `asType()` are not complete either! It must be a union of both.

Note: this does not require any javac flags and I tested it with Java 21.

Fixes #111

### DIFF
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -567,12 +567,12 @@ class InternalRecordBuilderProcessor {
     }
 
     private boolean isNotNullAnnotated(RecordClassType component) {
-        return component.getCanonicalConstructorAnnotations().stream().anyMatch(annotation -> notNullPattern
+        return component.getAccessorAnnotations().stream().anyMatch(annotation -> notNullPattern
                 .matcher(annotation.getAnnotationType().asElement().getSimpleName().toString()).matches());
     }
 
     private boolean isNullableAnnotated(RecordClassType component) {
-        return component.getCanonicalConstructorAnnotations().stream().anyMatch(annotation -> nullablePattern
+        return component.getAccessorAnnotations().stream().anyMatch(annotation -> nullablePattern
                 .matcher(annotation.getAnnotationType().asElement().getSimpleName().toString()).matches());
     }
 


mattbertolini:
I agree there is something off. My hunch is it's a JavaPoet thing writing the annotation in the wrong thing but I've been unable to build a reproducer for that project to isolate the problem. All the record builder can do is provide the annotations it has and I _think_ that's what this PR is doing now.

mattbertolini:
Nice! So I did finally figure out how to get JavaPoet to place the type use annotation in the right place! JavaPoet wants you to add the annotation to the `TypeName`. So in the `ElementUtils.getRecordClassType` when the `TypeName` is being constructed for each `RecordComponent`, we could add all TYPE_USE annotations there. I did a quick hack by adding all annotations onto the `TypeName` on my local copy to try and see if it would work.

```java
public static RecordClassType getRecordClassType(ProcessingEnvironment processingEnv,
            RecordComponentElement recordComponent, List<? extends AnnotationMirror> accessorAnnotations,
            List<? extends AnnotationMirror> canonicalConstructorAnnotations) {
    // I'm adding all annotations here but we might be able to just add type use annotations
    List<AnnotationSpec> annotationSpecs = accessorAnnotations.stream().map(AnnotationSpec::get)
            .filter(annotationSpec -> !annotationSpec.type().equals(TypeName.get(Override.class))).toList();
    // Notice the extra annotated method adding all the additional annotations
    var typeName = TypeName.get(recordComponent.asType()).annotated(annotationSpecs);
    var rawTypeName = TypeName.get(processingEnv.getTypeUtils().erasure(recordComponent.asType()));
    return new RecordClassType(typeName, rawTypeName, recordComponent.getSimpleName().toString(),
            recordComponent.getSimpleName().toString(), accessorAnnotations, canonicalConstructorAnnotations);
}
```

Adding all annotations duplicates the annotations in many cases so it's probably not ideal but I disabled adding annotations everywhere else for my test just to see if it works. It does! The Jspecify type_use annotation was exactly where I was expecting it! 

I'm not sure how best to introduce this in a more robust fashion but I wanted to let you know my findings. I'm happy to open a new issue or discussion up if you want to chat further. I think it may make sense to split out the type_use annotations from the accessor and constructor annotations and pass them into `getRecordClassType`. Then the annotations can be placed onto the type where necessary. What do you think?

facboy:
FYI `50` has broken compilation for me with `@Nullable` on inner classes, for the use cases above. I've had to roll back to 49.

facboy:
possibly fixes 'missing' `TYPE_USE` annotations? #260

PR #254, opened 2025-10-24
[#253] option addGeneratedAnnotation for RecordBuilder

### LINKED ISSUE #253: Make @Generated optional on generated classes
Hi!

The `@Generated` annotation causes some troubles with JPMS. 

## History on `@Generated`

Originally, it was designed to show static code analyzers which classes to ignore. Nowadays, they will usually look for `.gitignore`d files or something line `target/generated-*` or any `@*Generated` annotation. So that's that.

## Why it is a problem with JPMS

I do not want code which uses my record builders to require the `java.compiler` module -- it is not needed, because none of that code would read the `@Generated` annotation. So, instead of using `requires java.compiler;`, the correct approach is `requires static java.compiler` -- which translates to "only use this module for compiling".

However, the compiler is a bit dumb, sees the `@Generated` annotation and complains that it will be there on the `.class` file -- which it will be not, because it does not know about annotations and their `@RetentionPolicy`, so it assumes it will be there.

This will yield the following compiler warning:

```
[WARNING] /path/to/my/target/generated-sources/annotations/MyRecordBuilder.java:[217,6]  class javax.annotation.processing.Generated in module java.compiler is not indirectly exported using 'requires transitive'
```

### 1.: Do not generate `@Generated` at all

An option with all its drawbacks

### 2.: Only add `@Generated` if it is in the classpath

mapstruct uses this approach

### 3.: let the user decide (boolean)

Booleans are meh. Use an enum.

### 4. let the user decide (enum)

`-AgeneratedAnnotation` = `enum AddGeneratedAnnotation { ON, OFF, USER_SUPPLIED, IF_ON_CLASSPATH }`

Default: `IF_ON_CLASSPATH`

When `USER_SUPPLIED` is set, the user would also need to specify `-AgeneratedUserAnnotation=com.my.project.MyProjectGenerated` or so.

names are not meant to be final. Let some native speaker pick something better :)

bmarwell:
Refs: #253
Fixes: #253 

TODO:

- [ ] Implement for InterfaceBuilder

### DIFF
--- a/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
+++ b/record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
@@ -268,6 +268,16 @@ public @interface RecordBuilder {
         String beanClassName() default "";
 
         /**
+         * When to add the {@link javax.annotation.processing.Generated} annotation to generated files
+         *
+         * <p>
+         * If set to 'ALWAYS' (default), then always add the {@link javax.annotation.processing.Generated} annotation to
+         * generated files.
+         * </p>
+         */
+        GeneratedAnnotation addGeneratedAnnotation() default GeneratedAnnotation.ALWAYS;
+
+        /**
          * If true, generated classes are annotated with {@code RecordBuilderGenerated} which has a retention policy of
          * {@code CLASS}. This ensures that analyzers such as Jacoco will ignore the generated class.
          */
@@ -376,6 +386,18 @@ public @interface RecordBuilder {
 
     enum ConcreteSettersForOptionalMode {
         DISABLED, ENABLED, ENABLED_WITH_NULLABLE_ANNOTATION,
+    }
+
+    enum GeneratedAnnotation {
+        /**
+         * Always add the {@link javax.annotation.processing.Generated} annotation (default).
+         */
+        ALWAYS,
+        /**
+         * Never add the {@link javax.annotation.processing.Generated} annotation.
+         */
+        NEVER,
+        /* IF_ON_CLASSPATH, */ /* USER_SUPPLIED, */
     }
 
     /**
--- a/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
+++ b/record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
@@ -37,8 +37,7 @@ import static io.soabase.recordbuilder.processor.CollectionBuilderUtils.SingleIt
 import static io.soabase.recordbuilder.processor.CollectionBuilderUtils.SingleItemsMetaDataMode.STANDARD_FOR_SETTER;
 import static io.soabase.recordbuilder.processor.ElementUtils.getClassTypeFromNames;
 import static io.soabase.recordbuilder.processor.ElementUtils.getWithMethodName;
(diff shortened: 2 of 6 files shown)


record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -267,6 +267,16 @@
...
 
+        /**
+         * When to add the {@link javax.annotation.processing.Generated} annotation to generated files
+         *
+         * <p>
+         * If set to 'ALWAYS' (default), then always add the {@link javax.annotation.processing.Generated} annotation to
+         * generated files.
+         * </p>
+         */
+        GeneratedAnnotation addGeneratedAnnotation() default GeneratedAnnotation.ALWAYS;
```
Randgalt: Let's use a boolean here instead.

bmarwell: So, not giving a choice for later options, like `IF_ON_CLASSPATH` or `USE_USER_SUPPLIED_ANNOTATION`? Just want to check. Booleans are restrictive, no choice of extending the options later on!

Randgalt: We already have an option for "if on classpath" (that should be respected, BTW) so that ship has already sailed. Also, the use of the enum is very cumbersome. Let's use a boolean please.

bmarwell: I haven't seen such an option. How can I enabled the "Generated" annotation already? It's not on the classpath and it got generated.
If we could make that working, let's just drop by this PR and issue.

bmarwell: @Randgalt I have looked for this option but did not find it.

record-builder-core/src/main/java/io/soabase/recordbuilder/core/RecordBuilder.java
```diff
@@ -378,6 +388,18 @@ enum ConcreteSettersForOptionalMode {
         DISABLED, ENABLED, ENABLED_WITH_NULLABLE_ANNOTATION,
     }
 
+    enum GeneratedAnnotation {
```
Randgalt: If we use this enum I'd like to think of better names. `ALWAYS` and `NEVER` are limiting. We will have trouble add another values that works with these. Some alternatives:

- `ENABLED` / `DISABLED`
- `INCLUDE` / `EXCLUDE`

Also, we could name the method `generatedAnnotationMode` or something similar.

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/InternalRecordBuilderProcessor.java
```diff
@@ -80,11 +79,16 @@ class InternalRecordBuilderProcessor {
...
+        builder = TypeSpec.classBuilder(builderClassType.name()).addModifiers(metaData.builderClassModifiers())
+                .addTypeVariables(typeVariables);
         if (metaData.addClassRetainedGenerated()) {
             builder.addAnnotation(recordBuilderGeneratedAnnotation);
         }
+        switch (metaData.addGeneratedAnnotation()) {
+        case ALWAYS -> builder.addAnnotation(generatedRecordBuilderAnnotation);
+        case NEVER -> {
+            /* no-op */ }
+        }
```
Randgalt: Let's move this to a utility as it's always the same. The utility might have ctors/factories for different builder types.

PR #260, opened 2025-11-06
Add `TYPE_USE` annotations where possible

facboy:

PR #261, opened 2025-11-07
Add TYPE_USE annotations where possible

Randgalt:
Added one commit to https://github.com/Randgalt/record-builder/pull/260

@facboy you pushed from `master`. Next time please create a branch.


Randgalt:
Thank you for the contribution @facboy

CfFanDuel:
Is it possible to get this released? :)

jspecify annotations for me are now in a middle ground with staged builders where the @nullable annotation is correctly causing the stage not to be required, but as the annotation isn't being copied across to the builder I get errors when trying to pass in a nullable value.

This is the final piece of the puzzle :)

Randgalt:
done

PR #265, opened 2025-12-05
Allow @DeconstructorAccessor on fields

### LINKED ISSUE #264: Deconstructor Lombok integration
The usage of `@Getter`, `@Data`, `@Value` generate a getter method, and that method can't be used in combination with `@Deconstructor`

A use case example
```java
@Deconstructor
@Getter
public class MyClass {
    @DeconstructorAccessor
    private final int qty;

    @DeconstructorAccessor
    private final String name;
    
    private MyClass(int qty, String name) {
        this.qty = qty;
        this.name = name;
    }    
    // etc.
}
```

Note, Lombok has support for fluent accessors that will omit the `get` prefix.

Randgalt:
For uses cases such as Lombok, allow `@DeconstructorAccessor` on fields and search for a matching getter to use as the accessor.

Additionally, refine ordering rules for generated records. The default
is now encounter order.

Closes #264

cc @Srdjan-V - please test

### DIFF
--- a/deconstructors.md
+++ b/deconstructors.md
@@ -24,7 +24,9 @@ A deconstructor is either:
     - is annotated with `@RecordBuilder.Deconstructor`
 2. a class or interface that is:
    - annotated with `@RecordBuilder.Deconstructor`
-   - has one or more accessor methods annotated with `@RecordBuilder.DeconstructorAccessor`
+   - has one or more accessor methods or fields annotated with `@RecordBuilder.DeconstructorAccessor`. If it is
+   a field that is annotated, there must be a matching public, non-static method that is named either the same as the field
+   or is prefixed with `get` or `is`.
 
 When RecordBuilder encounters a deconstructor it generates a `record` that matches the deconstructor's parameters
 and, optionally, creates a record builder for it.
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
         <javax-el-version>3.0.1-b09</javax-el-version>
         <central-publishing-maven-plugin-version>0.7.0</central-publishing-maven-plugin-version>
         <jspecify-version>1.0.0</jspecify-version>
+        <lombok-version>1.18.42</lombok-version>
     </properties>
 
     <name>Record Builder</name>
@@ -176,6 +177,12 @@
                 <groupId>org.jspecify</groupId>
                 <artifactId>jspecify</artifactId>
                 <version>${jspecify-version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok-version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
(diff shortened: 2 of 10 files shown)


Srdjan-V:
@Randgalt This will only work if the get method is hardcoded, it will not catch the method that will be generated.

image


image

Randgalt:
@Srdjan-V how do I create an example that has "generated" methods?

Srdjan-V:
@Randgalt you'd have to decompile the bytecode.

```java
import lombok.Getter;

@Getter
public class MyClass {
    private final int qty;
    private final String name;

    private MyClass(int qty, String name) {
        this.qty = qty;
        this.name = name;
    }
    // etc.
}
```

```java
//
// Source code recreated from a .class file by IntelliJ IDEA
// (powered by FernFlower decompiler)
//

import lombok.Generated;

public class MyClass {
    private final int qty;
    private final String name;

    private MyClass(int qty, String name) {
        this.qty = qty;
        this.name = name;
    }

    @Generated
    public int getQty() {
        return this.qty;
    }

    @Generated
    public String getName() {
        return this.name;
    }
}
```

Randgalt:
Hmm - I'm not following. I have tests in this PR that work correctly and find the generated methods. What's wrong here?

Srdjan-V:
I'm guessing that the Deconstructor processor is running before the Lombok one in my case. I'm also using Gradle and its plugin for Lombok.

Randgalt:
I changed the order of processors and it fails as you see. However, it's not possible to access `Symbol.ClassSymbol` - those are protected classes. It doesn't seem too much to require that lombok run before record-builder so that record-builder can see the modifications. Unless we can find some other way.

Srdjan-V:
From what I can tell, `typeElement.getAnnotationMirrors()` will also give you the same information about annotations as `Symbol.ClassSymbol`

And it's not too much to require, but I will need to investigate why it's doing it in this order. I will write once I find out more.

Srdjan-V:
The plugin was using `extendsFrom` to add the Lombok processor, and that is not adding first. Explicitly adding it first made it work.

Randgalt:
@Srdjan-V shall we merge this then?

Srdjan-V:
Yeah it's fine. But It would also be nice for the deconstructor to auto find the methods, that might be to much though.

Randgalt:
> It would also be nice for the deconstructor to auto find the methods

I'm not sure how. Any ideas?

Srdjan-V:
It has the logic of finding methods, adding an option to the `@Deconstructor` to treat all fields as if they had the `@DeconstructorAccessor` should work.

```java
@Deconstructor(autoFindAccessors = true)
@Getter
public class MyClass {
    private final int qty;
    private final String name;
    
    private MyClass(int qty, String name) {
        this.qty = qty;
        this.name = name;
    }    
    // etc.
}
```

Randgalt:
The problem is naming. We'd need to add additional settings to manage `getFoo` vs `foo` vs `isFoo`. My initial reaction is that this is too much for this library when there's a simple workaround.

Srdjan-V:
No problem. This can be merged.

PR #276, opened 2026-06-20
Start claiming the annotations we process + New option `inheritRecordAnnotations`

### LINKED ISSUE #273: Add `@SuppressWarnings` to all generated classes
Hi,

Thanks for a great library!

We hit quite an unfortunate situation. We added `@Deprecated` to some components of our records and generated builders can't be compiled now because of the absence of `@SuppressWarnings`.

It can be a narrow `@SuppressWarnings("deprecation")`. But actually, I'd suggest adding `@SuppressWarnings({"all", "cast"})` to all generated classes. Generated code is not supposed to be checked by static analysis.

PS: Adding `cast` to `all` because `all` doesn't include it. Take a look: https://github.com/projectlombok/lombok/issues/1363

Randgalt:
Augments `inheritComponentAnnotations` by supplying a list of record-level
annotations that should be copied. Nominally this is for `SuppressWarnings`
but it can be used for any set of annotations.

Fixes https://github.com/Randgalt/record-builder/issues/273

### DIFF
--- a/options.md
+++ b/options.md
@@ -56,20 +56,20 @@ The names used for generated methods, classes, etc. can be changed via the follo
 
 ## Miscellaneous
 
-| option                                                                     | details                                                                                                                                                          |
-|----------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `@RecordBuilder.Options(inheritComponentAnnotations = true/false)`         | If true, any annotations (if applicable) on record components are copied to the builder methods. The default is `true`.                                          |
-| `@RecordBuilder.Options(publicBuilderConstructors = true/false)`           | Makes the generated builder's constructors public. The default is `false`.                                                                                       |
-| `@RecordBuilder.Options(builderClassModifiers = {}})`                      | Any additional `javax.lang.model.element.Modifier` you wish to apply to the builder.                                                                             |
-| `@RecordBuilder.Options(beanClassName = "Foo")`                            | If set, the Builder will contain an internal interface with this name.                                                                                           |
-| `@RecordBuilder.Options(addClassRetainedGenerated = true/false)`           | If true, generated classes are annotated with `RecordBuilderGenerated`. The default is `false`.                                                                  |
-| `@RecordBuilder.Options(addStaticBuilder = true/false)`                    | If true, a functional-style builder is added so that record instances can be instantiated without `new()`. The default is `true`.                                |
-| `@RecordBuilder.Options(inheritComponentAnnotations = true/false)`         | If true, any annotations (if applicable) on record components are copied to the builder methods. The default is `true`.                                          |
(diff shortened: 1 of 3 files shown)


Randgalt:
attn: @remal and @semyon-levin-workato

record-builder-processor/src/main/java/io/soabase/recordbuilder/processor/RecordBuilderProcessor.java
```diff
@@ -56,13 +56,14 @@ public class RecordBuilderProcessor extends AbstractProcessor {
...
     private final Set<TypeName> createdDeconstructors = ConcurrentHashMap.newKeySet();
+    private boolean annotationWasProcessed;
 
     @Override
     public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
         annotations.stream().filter(annotation -> annotation.getKind() == ElementKind.ANNOTATION_TYPE)
                 .forEach(annotation -> roundEnv.getElementsAnnotatedWith(annotation)
                         .forEach(element -> process(annotation, element)));
-        return false;
+        return annotationWasProcessed;
```
semyon-levin-workato: `getSupportedAnnotationTypes()` returns `"*"`, so when `process()` returns `true`, javac removes all annotations from the processing round. Subsequent processors (Immutables in my case) never see their annotations.

From `JavacProcessingEnvironment.java` (JDK 24):
```java
// line 860: "*" matches every annotation, so matchedNames = all annotations in the round
for (Map.Entry<String, TypeElement> entry: unmatchedAnnotations.entrySet()) {
    if (ps.annotationSupported(unmatchedAnnotationName)) {
        matchedNames.add(unmatchedAnnotationName);
    }
}

// line 882: process() returned true -> remove all matched names
if (processingResult) {
    unmatchedAnnotations.keySet().removeAll(matchedNames);
}
// unmatchedAnnotations is now empty, loop exits, no more processors run
```

Either revert to `return false` or narrow `getSupportedAnnotationTypes()` to the specific annotations record-builder handles.

Randgalt: Yeah, I was unsure about that part. I'll remove it.

Randgalt: @semyon-levin-workato I've removed it. Try now.

semyon-levin-workato: works good

semyon-levin-workato (approved):
@Randgalt , the solution works for me. What would you suggest we do next?
1. I close my PR
2. You release this one
Do I get this right?

Randgalt:
Yes, I'll do that.

Randgalt:
v53 is released

